PR #1, opened 2022-06-12
formula exec: dynamically follow and mount symlinks for certs in fetch container

Kirth:
small PR to dynamically follow all symlinks in the /etc/ssl/certs directory and mount them so that they may resolve in the exec/guest container.  Use case is running warpforge on NixOS, which uses multiple layers of symlinks there.

not sure if the helper method is named appropriately or in the correct spot

### DIFF
--- a/pkg/formulaexec/formula_exec.go
+++ b/pkg/formulaexec/formula_exec.go
@@ -83,6 +83,11 @@ func getMountDirSymlinks(start string) []string {
 	fi, _ := os.Lstat(path)
 	for fi != nil && fi.Mode() & os.ModeSymlink == os.ModeSymlink {
 		pointee, _ := os.Readlink(path)
+		if !filepath.IsAbs(pointee) {
+			// the symlink points to a relative path
+			//add the symlink's path to create an absolute path
+			pointee = filepath.Join(filepath.Dir(path), pointee)
+		}
 		path = pointee
 		paths = append(paths, path)
 		fi, _ = os.Lstat(path)
@@ -123,11 +128,6 @@ func getNetworkMounts(wsPath string) []specs.Mount {
 
 			for _, p := range getMountDirSymlinks(path) {
 				dir := filepath.Dir(p)
-
-				// ignore relative symlinks
-				if !filepath.IsAbs(p) {
-					continue
-				}
 
 				// ignore duplicate mounts
 				duplicate := false


warpfork:
Alright, tl;dr, I think we could probably merge this, and might as well -- the code isn't too much added complexity, and it makes a thing work that doesn't work right now.

But there's also reasons for blinking warily at this, and I'm going to spend the rest of this comment talking about why, and how we got in the entire situation here, because I can foresee this code zone coming up again in the future.

---


**We never wanted to be owning anything about certificate management in Warpforge.**

So how did we end up here at all?

## Backstory

### We do filesystem provisioning in a container.

Why?  (This really is a "five why's" sort of journey...)

Because we want Warpforge to run in userland -- without root permissions on the host.

Okay, why's that matter?  Because: userland containers requires using some features of containers called "uid mapping".

Doing all of our filesystem provisioning work in a container means that we can use {whatever the uid mapping config is for the real task} in the provisioning container too, and thus, all the numbers will line up, and in a very foolproof way.  (Replicating the semantics without using the container system might be possible, but it seemed like asking for trouble -- and honestly, isolating the I/O processes is neat anyway.)

### ... which means there's an allowlist of inputs to the provisioning job.

Which is generally good.  We like hermetic, right?!  That's like, our whole and entire jam, here.

Yeah, well.  It's not critical, in this particular situation.  We're unpacking the data (using the uid mappings!) but also fetching the dang stuff in the first place.  So, necessarily, we're okay with a pretty drastic drop in hermeticism _for this provisioning container_, because at minimum, we're going to need to have network access enabled -- not much fetching is gonna get done without that!

And a lot of fetching, in practice, uses "`https://`".  This is what upgrades the hermeticism from being "not critical" to "actually, kind problematic" in this case.

### "https" means "certificate problems"

"https://" connections need a certificate store on the host to check the "s" part against.

It's not correct to hardcode the certificate store for a variety of reasons.  Most dominantly, and "good game" alone, is that _certificates expire_.  They're time limited.  (Root certs certainly don't expire often, but they do.  Time marches on.)

So we need to get certificates into the provisioning environment, somehow.  And the ones on the host... might as well do.

### so... mounts?  mounts.

It looked (at first pass; and maybe we're going to want to revisit this at some point) like simply mounting the certificate directory from the host into the provisioning container would solve it.

One mount (aah; now we're seeing this isn't true) is not a huge problem; it's a drop in hermeticism, but as discussed above, for the provisioning container, that's fine; okay, sure.  (Also, this would probably only work on linux and mac; windows has a whole Something Else going on; but that's a bridge to cross later, somehow, anyway.)

---

## ... this isn't working very well.

I mean.  It's working.  Like I said, this patch does "lgtm".

But we're finding a lot of complexity here.  More than I'm happy about.  It's now an unknown number of mounts; and there's a bunch of work being done to scan the filesystem for symlinks, which is... okay, but it's probably slowing down our launch of these provisioning containers, however minorly.  Just at a super empirical level, this has rapidly turned out to be our biggest portability problem: every new user we've had on a new distro has required a new patch to this fracas.  I hate portability problems.  And this is a very dumb one, being from managing of things I'd rather not manage at all.

If anyone out there reading this, whether now or a year from now, can see a simpler cut through this whole space, I'd be very, very happy to hear ideas.

willscott:
> But we're finding a lot of complexity here.

I wonder if it's worth dealing with the complexity in a single virtual filesystem mount - e.g. having the mount seen by userspace backed by nfs/p9/fuse/etc to reason about the exposed file system with application-level logic

warpfork:
Maybe?  I'm not sure it would be winning in any currencies I value, though.

I did a back-of-the-envelope check on if 100 mounts is more expensive than 1, and if 200 is noticeably more expensive than 100.  It didn't seem actively problematic to encounter higher numbers.

Mostly I'm concerned about owning the complexity of it at all.  So far we've hit special cases for Arch, and another set of more expansive special cases for Nix -- the special-case-per-distro ratio is darn near 1:1 and I don't care for that at all.  Moving the logic between {preparatory inspection that generates mount configs} and {logic backing a programmatic mount in realtime} doesn't really reduce complexity, afaict; just shifts it from one place to another.

warpfork:
Other things that cross my mind would include "what if we separate {fetch} from {unpack}?" -- because we only care about the uid mappings during unpack -- maybe doing fetch without any sandboxing or containerization at all would "free" us from this problem (not fixing it, but making it less ours to own).

However, that's overall still sort of playing a shell game to displace the complexity; for some packing strategies and transport strategies, implementing to that described separation is easy; in others, not so much.

pkg/formulaexec/formula_exec.go
```diff
@@ -98,18 +113,29 @@ func getNetworkMounts(wsPath string) []specs.Mount {
...
+	// some distros use symlinks in /etc/ssl/certificates
+	// if this directory exists, mount and follow all symlinks
+	// so that they will resolve.
+	if certFiles, err := ioutil.ReadDir("/etc/ssl/certs"); err == nil {
+		for _, file := range certFiles {
+			path := filepath.Join("/etc/ssl/certs", file.Name())
+
+			for _, p := range getMountDirSymlinks(path) {
+				dir := filepath.Dir(p)
+				fmt.Println(dir)
```
lukegb, line 126: Some debugging code that got left in?

Kirth: Yep!  Updated my commit.

warpfork: Thank you for the additional eyes! :D

warpfork:
@ericevenchick do you wanna give this a quick eyeball, and if no objections, merge?

I'm thinking that if it's running fine for Kirth on NixOS, and it still runs in our environment on github actions, it seems like it's not going to be a functional regression, at least.

warpfork:
Per conversation in call with @ericevenchick right now as we look at this again: I think what this should probably do is actually:

- for each file in the certs dir: if it is a symlink:
- read the symlink target string; if it's absolute, proceed; if it's not absolute, `filepath.Join` with the current dir (which results in an absolute path again).
- chop the filename off that path, and add that dir to the mount set (if not already present);
- if the path (before chopping) is a symlink: repeat from step 2.

This turns all paths into absolute proactively, which produces sane results and is easy to dedup; and accumulates all directories relevant along the way if symlinks are chained.

warpfork:
(I pinged @Kirth elsewhere to test, and he confirmed it still works on his machine after the changes we heaped on!  Woohoo!)

PR #2, opened 2022-07-28
Add capsule types

ericevenchick:
Add "Capsule" types for various types in wfapi. These Capsules allow us to version the type by using an IPLD union, and create breaking changes in the future without losing backwards compatibility. This also fixes all tests and other code which now needs to support the Capsule types.

### DIFF
--- a/examples/100-formula-parse/example-formulas.md
+++ b/examples/100-formula-parse/example-formulas.md
@@ -45,13 +45,15 @@ Here's about the simplest formula one could have:
 ```json
 {
 	"formula": {
-		"inputs": {},
-		"action": {
-			"exec": {
-				"command": []
-			}
-		},
-		"outputs": {}
+		"formula.v1": {
+			"inputs": {},
+			"action": {
+				"exec": {
+					"command": []
+				}
+			},
+			"outputs": {}
+		}
 	}
 }
 ```
@@ -80,16 +82,18 @@ Formulas usually have inputs.  And the most common type is a "ware", which will 
 ```json
 {
 	"formula": {
-		"inputs": {
-			"/mount/path": "ware:tar:qwerasdf",
-			"/other/place": "ware:git:abcd1234"
-		},
-		"action": {
-			"exec": {
-				"command": []
-			}
-		},
-		"outputs": {}
+		"formula.v1": {
+			"inputs": {
+				"/mount/path": "ware:tar:qwerasdf",
+				"/other/place": "ware:git:abcd1234"
+			},
+			"action": {
+				"exec": {
+					"command": []
+				}
+			},
+			"outputs": {}
+		}
 	}
 }
 ```
@@ -103,17 +107,19 @@ And in this case, they also are more likely to use a "literal" form.
 ```json
 {
 	"formula": {
-		"inputs": {
-			"$USER": "literal:hello",
-			"$HOME": "literal:/home/hello",
-			"$PATH": "literal:/bin:/usr/bin:/local/bin"
-		},
-		"action": {
-			"exec": {
-				"command": []
-			}
-		},
-		"outputs": {}
+		"formula.v1": {
+			"inputs": {
+				"$USER": "literal:hello",
+				"$HOME": "literal:/home/hello",
+				"$PATH": "literal:/bin:/usr/bin:/local/bin"
+			},
+			"action": {
+				"exec": {
+					"command": []
+				}
+			},
+			"outputs": {}
+		}
 	}
 }
 ```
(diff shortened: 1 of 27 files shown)


warpfork:
Nice.

warpfork:
So: pretty much every serial API document has the indentation drift one step further to the right... but in exchange, we become drastically less worried about versioning and clarity of any migrations in the entirety of the future.

Worthy trade :+1: :+1: :+1:

examples/100-formula-parse/example-formulas.md
```diff
@@ -161,7 +169,7 @@ and what types we see after parsing the information:
 [testmark]:# (hello-and-debug/formula.debug)
 ```text
 struct<FormulaAndContext>{
-	formula: struct<Formula>{
+	formula: union<FormulaCapsule>{struct<Formula>{
```
warpfork, line 172: It's kinda neat how this debug printout syntax collapses the union/variant into zero additional lines nor indentation.

Kinda wish we could get the JSON output to do that too.  I bet it would make the JSON look a lot prettier.

pkg/formulaexec/formula_exec.go
```diff
@@ -579,16 +578,20 @@ func GetBinPath() (string, wfapi.Error) {
...
-	context := fc.Context
 	rr := wfapi.RunRecord{}
 
-	// convert formula to node
-	nFormulaAndContext, errRaw := bindnode.Wrap(&fc, wfapi.TypeSystem.TypeByName("FormulaAndContext")).LookupByString("formula")
-	if errRaw != nil {
-		// panic! this should never fail unless the TypeSystem is broken
-		panic(fmt.Sprintf("Fatal IPLD Error: bindnode.Wrap failed for FormulaAndContext: %s", errRaw))
+	if fc.Formula.Formula == nil {
+		return rr, wfapi.ErrorFormulaInvalid("no v1 Formula in FormulaCapsule")
```
warpfork, line 584: Hm, maybe this should say something about "unrecognized version" in the message?

Might be worth a different error code too?  Less sure on that -- at the moment it won't make much practical difference.  But it seems to me there's a semiotic difference between "i tried to parse this and it violated some constraints" and "i tried to parse this and it clearly wasn't a message version that I understand".

warpfork: Hang on.  On second thought, I might be blowing smoke here.  Is this branch even going to do anything right now?  If this thing is nil... then the schema parse should've flunked this data out already.  This branch should be unreachable, until the schema actually gets more than one member in that capsule type.

pkg/formulaexec/formula_exec.go
```diff
@@ -599,14 +602,14 @@ func execFormula(ws *workspace.Workspace, fc wfapi.FormulaAndContext, formulaCon
 		Codec:    0x71, // 0x71 means "dag-cbor" -- See the multicodecs table: https://github.com/multiformats/multicodec/
 		MhType:   0x20, // 0x20 means "sha2-384" -- See the multicodecs table: https://github.com/multiformats/multicodec/
 		MhLength: 48,   // sha2-384 hash has a 48-byte sum.
-	}}, nFormulaAndContext.(schema.TypedNode).Representation())
```
warpfork, line 602: I see a somewhat critical little variable naming typofix creeps in here as well... ;P

wfapi/wfapi.ipldsch
```diff
@@ -337,13 +359,21 @@ type ItemLabel string
...
 
-type CatalogMirrorByWare {WareID:WarehouseAddrList}
-type CatalogMirrorByModule {ModuleName:CatalogMirrorsByPacktype}
+type CatalogMirrorsByWare {WareID:WarehouseAddrList}
+type CatalogMirrorsByModule {ModuleName:CatalogMirrorsByPacktype}
 type CatalogMirrorsByPacktype {Packtype:WarehouseAddrList}
+
+type ReplayCapsule union {
+	| Plot "plot.v1"
+} representation keyed
```
warpfork, line 379: whitespace police

wfapi/wfapi.ipldsch
```diff
@@ -337,13 +359,21 @@ type ItemLabel string
...
 
 type WarehouseAddrList [WarehouseAddr]
 
-type CatalogMirrorByWare {WareID:WarehouseAddrList}
-type CatalogMirrorByModule {ModuleName:CatalogMirrorsByPacktype}
+type CatalogMirrorsByWare {WareID:WarehouseAddrList}
+type CatalogMirrorsByModule {ModuleName:CatalogMirrorsByPacktype}
 type CatalogMirrorsByPacktype {Packtype:WarehouseAddrList}
+
+type ReplayCapsule union {
```
warpfork, line 377: :thinking: Okay, thinking outloud for a moment:

This type isn't strictly necessary, I think?  We could just use PlotCapsule in the code relating to replays, and that would be functionally correct?

But writing it this way with the extra type can be argued to be clearer; and if we do introduce new versions of replay formation that begin to contain more than just a plot, then this type would become necessary anyway.  Putting it in now will make such a future evolution slightly more obvious to implement.  (And we have a niggling suspicion we might do such a thing.)

Yeah, okay.  Seems legit.

warpfork (approved):
LGTM!

PR #3, opened 2022-08-02
Update catalog fs projection

ericevenchick:
Change the `releases` and `replays` directories in the catalog to `module.releases` and `module.replays`. This ensures that we do not end up with a naming conflict if someone uses the words `releases` or `replays` in their catalog path.

Also, replays now use a content-addressable filename instead of the version name. This makes sense since they're addressed using an IPLD link.


warpfork (approved):
LGTM

ericevenchick:
After discussion, we're going with underscore prefixes instead of the `module.replays` and such.

warpfork:
LGTM, round 2.

And thanks to cbehling for asking more or less "hey, couldn't this be shorter" in the [warpforge chat on matrix](https://matrix.to/#/#warpforge:matrix.org), which got us here.

PR #6, opened 2022-08-09
Fix runpath being overridden

TripleDogDare:
Fixes silent overriding of the run path to an empty string
when temp dir creation failed.


warpfork:
Ow.

Thank you for finding this and fixing it!

PR #7, opened 2022-08-10
Workaround for terminal echo lock

TripleDogDare:
Adding stty sane after tests should reset the terminal echo settings and
anything else that may have been inadvertently corrupted. This should help
reduce the loss in productivity caused by this problem, however we would
like to find the root cause before it becomes an issue for users.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,7 @@ else
 	$(SERUM) -strict ./...
 endif
 	go test ./...
+	@stty sane
 
 all: test install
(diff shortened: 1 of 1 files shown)


warpfork:
It's definitely a workaround, alright.  But I have to agree, this makes life a bit less bad in the meanwhile.  LGTM?

TripleDogDare:
Good enough for me (GEFM).

PR #8, opened 2022-08-11
fix issues with adding replays, allow for overwriting of releases

ericevenchick:
The schema changes broke adding replays and releases (i.e., `warpforge catalog release` command). This had to do with the order that we were writing things, which has been corrected here.

Also, the `-f` flag on the catalog command now allows for overriding data when needed.

### DIFF
--- a/pkg/workspace/catalog.go
+++ b/pkg/workspace/catalog.go
@@ -361,7 +361,8 @@ func (cat *Catalog) AddItem(
 	if hasItem && !overwrite {
 		// item exists but overwrite not requested, error
 		return wfapi.ErrorCatalogAlreadyExists(releaseFilePath, ref.ItemName)
-	} else if !hasItem {
+	}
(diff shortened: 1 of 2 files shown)


pkg/workspace/catalog.go
```diff
@@ -355,13 +356,16 @@ func (cat *Catalog) AddItem(
...
 	}
 
-	// ensure the item does not already exist
+	// check if the item exists
 	_, hasItem := release.Items.Values[ref.ItemName]
-	if hasItem {
+	if hasItem && !overwrite {
+		// item exists but overwrite not requested, error
 		return wfapi.ErrorCatalogAlreadyExists(releaseFilePath, ref.ItemName)
+	} else if !hasItem {
```
TripleDogDare, line 364: The `else` isn't necessary because the preceding branch returns.

PR #10, opened 2022-08-17
plot flag for ferk

ericevenchick:
Allow 'ferking' on custom plot files.


warpfork:
This is a very very good feature :raised_hands:

PR #14, opened 2022-08-30
No Net Tests

ericevenchick:
This diff allows tests to run without using the internet.

This required a few changes. We run our tests with testmark, which creates a unique directory for each run and breaks caching. This is solved by copying files over before running cli tests, and pointing Warpforge at its git directory for catalog information. This is probably not the best solution, so *consider it a WIP*.

We also do away with `min.warpforge.io` as a catalog since we do not need Alpine anymore. Busybox is now the rootfs for all tests.

Some tests (`catalog update`) have been removed because they do not make sense without a network connection.

TODO still: update README.md to remove min.warpforge.io and Alpine Linux references.


.test-home/.gitignore
```diff
@@ -1 +1,5 @@
-.warpforge/*
+.warpforge/warehouse/*
+.warpforge/cache
+.warpforge/catalogs
+.warpforge/runc-root
+!.warpforge/warehouse/4z9/
```
warpfork, line 5: You can just one-time force add things.  So probably this isn't needed.  Unless you reeeeally want it, for some sense of explicitness.

cmd/warpforge/main_test.go
```diff
@@ -82,12 +78,30 @@ func cleanRunRecord(str string) string {
 	return strings.TrimSpace(str)
 }
 
-func execFn(args []string, stdin io.Reader, stdout, stderr io.Writer) (int, error) {
-	err := makeApp(stdin, stdout, stderr).Run(args)
-	if err != nil {
-		return 1, err
+func buildExecFn(projPath string) func(args []string, stdin io.Reader, stdout io.Writer, stderr io.Writer) (int, error) {
```
warpfork, line 81: I have a suspicion that having a closure factory like this is gonna make life worse when we try to get error analysis piped through here properly (since I ran into that problem in my PR last week).

Fair enough to leave that to fix when we come to it, though.

examples/110-formula-usage/example-formula-exec.md
```diff
@@ -39,9 +39,7 @@ This formula echos a value to stdout.
 	},
 	"context": {
 		"context.v1": {
-			"warehouses": {
-				"tar:57j2Ee9HEtDxRLE6uHA1xvmNB2LgqL3HeT5pCXr7EcXkjcoYiGHSBkFyKqQuHFyGPN": "https://dl-cdn.alpinelinux.org/alpine/v3.15/releases/x86_64/alpine-minirootfs-3.15.0-x86_64.tar.gz"
-			}
+			"warehouses": {}
```
warpfork, line 42: I think a little snippet of prose in this area to comment on this might be good.

Perhaps:

> Note that for the "warehouses" field to be empty is a little unusual -- usually, URLs are present there, which warpforge will use if it needs to fetch any inputs.  In this demo, it can be blank, because our test harness pre-populates caches in the test environment, so the tests work without any need for URLs here.

warpfork: ... actually, gosh, now that I think about it a second longer -- maybe we should actually put the URLs back in here, even.

This is the "examples" folder.

It's good for our tests to run without network... but these should still be examples ready for someone else to copy-paste.  That means warehouse address URLs should be textually present.

Our tests in CI should probably work no-net, but... a cache hit "happens" to accomplish that.

Maybe a stretch goal should be: we add some flags to the test harnesses that will also run them _without_ the no-net setup process, so as to do a load-bearing check that these URLs are in fact live.  We wouldn't run those tests when developing, but having them available as a health check would seem reasonable.

examples/500-cli/cli.md
```diff
@@ -421,92 +418,45 @@ warpforge catalog ls
...
-				"https://dl-cdn.alpinelinux.org/alpine/v3.15/releases/x86_64/alpine-minirootfs-3.15.0-x86_64.tar.gz"
+			"tar:4z9DCTxoKkStqXQRwtf9nimpfQQ36dbndDsAPCQgECfbXt3edanUrsVKCjE9TkX2v9": [
+				"file://.warpforge/warehouse/4z9/DCT/4z9DCTxoKkStqXQRwtf9nimpfQQ36dbndDsAPCQgECfbXt3edanUrsVKCjE9TkX2v9"
 			]
 		}
 	}
 }
 ```
 
-#### git
```
warpfork, line 463: Maybe let's keep these, but just figure out how to put a feature flag on these tests?

I don't know what the gopherean state of the art is for that... there's the `t.Short` predicate, but I'm not a huge fan of that because it's just one bool.  Env vars connected to a `t.Skip` would be an option.  Build tags are maybe also an option?  I don't have strong opinions.

warpfork:
I spammed a lot of comments, but broadly speaking, yes, sweet, ohgoshyes, thank you.

The biggest ask I have is that we figure out what to do with any tests that _do_ legitimately use network.  Those are "smoke tests" or something, certainly, and I very much want to be able to choose not to exercise them if I'm hacking on a plane or whatever... but we probably don't want to throw them in the ditch entirely, either.

warpfork:
... you also may have had a branch merging typo that caused the #15 PR to end up in here, heh.  (Would be nice to have history cleanliness, but if it's trouble... it's also only one file, so, I guess it doesn't have a huge impact on diff.)

examples/500-cli/cli.md
ericevenchick: For now, I have put them back in and removed the testmark line... so the example is still there but there but it will not run. Looking for a better solution to allow net/no-net options.

examples/110-formula-usage/example-formula-exec.md
ericevenchick: They have been readded everywhere *except* the CLI runformula example. That is the only example that actually required network. I have added some prose to explain.

Ideally, would be nice to have these behind a net-flag but for now this seems like an acceptable tradeoff.

cmd/warpforge/main_test.go
ericevenchick: Yeah.... you're probably not wrong but also seems like a future problem.

examples/500-cli/cli.md
ericevenchick: Now behind the `-testutil.offline` flag.

examples/110-formula-usage/example-formula-exec.md
ericevenchick: Now behind the `-testutil.offline` flag.

.test-home/.gitignore
warpfork, line 5: Don't really need, if it's a single asset in here and you added it (or `add -f`'d it) already.

Might be worth a comment either here or in a spare file in that directory _why_ something's there that's gitignored, though, if we're worried about the clarity of it.

cmd/warpforge/main_test.go
```diff
@@ -49,6 +50,21 @@ func testFile(t *testing.T, fileName string, workDir *string) {
...
+			}
+			// we want to run the contents of the `/net` dir
+			t.Run(dir.Name, func(t *testing.T) {
+				test := testexec.Tester{
+					ExecFn:   execFn,
+					Patches:  &patches,
+					AssertFn: assertFn,
+				}
+				test.Test(t, *dir.Children["net"])
+			})
```
warpfork, line 66: Surely you don't need to duplicate all this just to change what the function call is applied on?

cmd/warpforge/main_test.go
```diff
@@ -82,12 +94,29 @@ func cleanRunRecord(str string) string {
...
+		err := os.Setenv("WARPFORGE_PATH", filepath.Join(projPath, "plugins"))
+
+		// set up a root workspace in the testmark run directory
+		testmarkWd, err := os.Getwd()
+		if err != nil {
+			return 1, fmt.Errorf("failed to get testmark pwd: %s", err)
+		}
+		os.MkdirAll(filepath.Join(testmarkWd, ".warpforge/catalogs/default"), 0755)
+		copyCmd := exec.Command("cp", "--recursive", filepath.Join(projPath, ".warpforge", "catalog"), filepath.Join(testmarkWd, ".warpforge", "catalogs"))
+		copyCmd.Run()
```
warpfork, line 109: This is missing error handling, but also, I'd be a lot happier if we could do this without resorting to any exec's.  It seems out of place and unexpected.

cmd/warpforge/main_test.go
```diff
@@ -82,12 +94,29 @@ func cleanRunRecord(str string) string {
...
+		copyCmd = exec.Command("cp", "--recursive", filepath.Join(projPath, ".warpforge", "warehouse"), filepath.Join(testmarkWd, ".warpforge"))
+		copyCmd.Run()
```
warpfork, line 111: Can we get rid of this copy operation entirely by adding a feature where an env var like `WARPFORGE_EXTRA_WAREHOUSEADDR=file+ca:///over/here` will just let us get this from where it already resides in the filesystem?

cmd/warpforge/main_test.go
```diff
@@ -82,12 +94,29 @@ func cleanRunRecord(str string) string {
...
+		// override the path to required binaries
+		err := os.Setenv("WARPFORGE_PATH", filepath.Join(projPath, "plugins"))
+
+		// set up a root workspace in the testmark run directory
+		testmarkWd, err := os.Getwd()
+		if err != nil {
+			return 1, fmt.Errorf("failed to get testmark pwd: %s", err)
+		}
+		os.MkdirAll(filepath.Join(testmarkWd, ".warpforge/catalogs/default"), 0755)
+		copyCmd := exec.Command("cp", "--recursive", filepath.Join(projPath, ".warpforge", "catalog"), filepath.Join(testmarkWd, ".warpforge", "catalogs"))
```
warpfork, line 108: How many pieces of data are we moving around with this?  Are they textual, such that we could stick them in a data hunk in testmark, and thus not need this operation?

Still thinking about exactly how that would work, but... anything we can do to make this piece of setup less magical, and more legible in docs and such, would probably be a win.

pkg/testutil/flags.go
```diff
@@ -0,0 +1,5 @@
+package testutil
+
+import "flag"
+
+var FlagOffline = flag.Bool("testutil.offline", false, "Disable network usage in tests")
```
warpfork, line 5: imo `s/testutil.offline/nonet` or `s/testutil.offline/offline/`.

Bikeshed color.  But shorter is better.  And I don't really see the need for dotted prefixes.  We're not a library; we don't need to namespace ourselves in this.

.test-home/.gitignore
ericevenchick: This dir should have been removed. Oops.

cmd/warpforge/main_test.go
ericevenchick: Surely you are correct.

pkg/testutil/flags.go
ericevenchick: Ok, `-offline` it is.

cmd/warpforge/main_test.go
```diff
@@ -49,13 +50,24 @@ func testFile(t *testing.T, fileName string, workDir *string) {
...
-		t.Run(dir.Name, func(t *testing.T) {
+		testName := dir.Name
+		testDir := dir
+		if _, hasNetTests := dir.Children["net"]; hasNetTests {
+			if *testutil.FlagOffline {
+				t.Log("skipping test", dir.Name, "due to offline flag")
+				continue
+			}
+			// we want to run the contents of the `/net` dir
+			testName = dir.Name
```
warpfork, line 61: Doesn't do anything?  `dir` hasn't been changed.

ericevenchick:
It's probably worth noting that this PR also makes warpsys the default catalog. I think that's a sane decision that we wanted anyway, it's a bit odd to do it here, but it's done now.

PR #13, opened 2022-08-26
Extracting filesystem accesses; 'workspace inspect' command

warpfork:
I got busy without network for a while, so a heap of things come at once (the individual commits may end up more sane to review than the whole PR at once):

- A new package is introduced, to be the home of anything to do with filesystem and data loading (see big doc file);
- A copious amount of refactoring to use `fs.FS` in more places;
- And there is now a `workspace inspect` subcommand, which takes a look around to see all the modules it can see, and gives some brief reporting on their health (are all imports resolvable, etc).
- Extracted some stats counting features for Plots to be reusable to serve that (it's the same stuff as `wf status` is doing, just now in bulk).
- (rather importantly) fixed that Plot stat counting stuff to be correctly recursive, rather than forget to count things beyond the top level object!

Everything about the appearance of the `workspace inspect` command should be considered comically WIP, for 'tis terribly ugly, but I hope it's an interesting start in a direction.

PR #16, opened 2022-09-09
catalogsite: first pass

warpfork:
This PR contains a working checkpoint of a (admittedly still very skeletal) html generator to make a web-navigable view of the data in a catalog.

The overall idea of this is to... well, like I said: make web-navigable thing.  The filesystem projection of a catalog into a git repo viewed on github is the next-best-thing we've got so far; I'd like to have something prettier, more fitted to the purpose, and with more crosslinking.

This is still just an outline.  It contains copious hacks in the "test".  It's not production-ready.  It's not even wired up to the CLI yet.  Nonetheless, I hoist the PR (and might want to merge it, and keep working iteratively from here, for productivity... I already let it get stale and had to do a big rebase once).

~The exact filesystem layout may be subject to change. This is a rebase of an old prototype (from before warpforge v0.2); the exact filesystem I'd like to use nowaday should more closely mirror the paths of the catalog filesystem projection in JSON, just with different extensions (which is not what we have here).~  Now updated.

Everything else _also_ subject to change :)  There's no styling, etc.

Future work includes turning this into a real CLI command;
giving it parameters for which catalog(s?) to generate content from;
teaching it to render and link up certain pieces of metadata (especially, right now, Replays);
downstream of rendering Replays, we can crosslink their imports to other parts of the catalog;
and probably a lot more.

(Oh, and in the guts, I'd like to refactor this to use https://github.com/warpfork/go-fsx in the fairly immediate future.  Right now, there's an unholy mixture of `fs.FS` and `os.*` because of the lack of stdlib fs interfaces for write operations.  The go-fsx repo is intended to address that, but I haven't pulled it in yet.  This might end up being addressed as part of https://github.com/warpfork/warpforge/pull/13, though.)


Storken:
Cool, looks like a great foundation, how would we go about if we'd use styling-files, like for example simple .css files. The html-file would need to know the relative path to it, or we could actually at a generate-step compile the .css file into the <head> of the html, which would remove the need to host a .css somewhere locally. What do you think would be a good approach here?

warpfork:
I think just another `//go:embed` line, and then a function somewhere that copies that to the output dir; then the `(url "css.css")` function (same as is used to crosslink the html pages already) should DTRT for the path to it.

... I'll add a stub that does this real quick.

warpfork:
... done.

It's a tad redundant in the html (we could probably use more templating features to make reusable snippets for the header?), but that's more of a polish thing, and keeps it simple for now.

warpfork:
Work will continue in more branches :)  Merging the progress so far!

PR #17, opened 2022-09-18
Catalogsite

Storken:
- Created a js.js file to remember previous choice in dark/lightmode and to toggle checkbox if darkmode is selected from start
- Fixed some inline-styling
- Added a dark/lightmode toggle
- Added some initial colors
- Styling added to `catalogIndex.tmpl.html`


warpfork:
This looks great!

Fetched and checked it out locally.

I'm always wary of JS, but this JS looks good.  It's very minimal: it is only used to persist user preference on the colors.  That seems totally fine: a noscripter won't even miss it.  And the toggle itself still works as a basic checkbox, needing none of the JS to have its effect.  Very nice.

The CSS animation on the button is a beautiful piece of clockwork :)

PR #18, opened 2022-09-19
style module and release pages

Storken:
## Description

Style catalogModule.tmpl.html and catalogRelease.tmpl.html to make them easier to navigate.

## Tasks

- [x] Breadcrumbs for easy overview of where you have navigated (and good way to get back)
- [x] Dark / light-mode
- [x] Toggling between tabs of data, it's better to show one at a time so the user easier understands and knows what they are looking at. Especially if there's a lot of data to view.
- [x] Don't add any unnecessary javascript

## How has this been tested?

Specify the platform(s) on which this was tested:
- Brave
- Chrome
- Safari

### DIFF
--- a/pkg/cataloghtml/catalogIndex.tmpl.html
+++ b/pkg/cataloghtml/catalogIndex.tmpl.html
@@ -21,7 +21,10 @@ which is not available and rigged up as a convenient library yet.
 <input type="checkbox" id="dark-mode-checkbox" />
 <div class="container">
 	<div class="header">
-		<h1 style="display:inline">Catalog</h1>
+		<div>
+			<small>&nbsp;</small>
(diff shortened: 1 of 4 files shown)


ericevenchick (approved):
lgtm :ship:

warpfork (approved):
I might have some nits to pick with the exact content of the breadcrumb links, but I'm not entirely sure what they are yet, and it does look good and perform the functions :)

Storken:
> I might have some nits to pick with the exact content of the breadcrumb links, but I'm not entirely sure what they are yet, and it does look good and perform the functions :)

Absolutely, this is just the first iteration :) I think this is much cleaner than the way we showed the links before though.

PR #20, opened 2022-09-27
Add cli for catalog html generation

ericevenchick:
Add a CLI for generating catalog html output.

By default, this will generate the output for the catalog `default` and place it into the `_html` subdirectory of that catalog. Flags can be used to change the catalog used, output directory and URL prefix:

`wf catalog --name=warpsys generate-html --output=/tmp/warpsys-catalog --url-prefix=https://example.com/some/prefix`

We may decide to only use relative paths and remove the need for `--url-prefix`.

### DIFF
--- a/cmd/warpforge/catalog.go
+++ b/cmd/warpforge/catalog.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bytes"
+	"context"
 	"fmt"
 	"os"
 	"os/exec"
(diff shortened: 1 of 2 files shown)


warpfork (approved):
Looks pretty fine to me!

The tests aren't doing a ton beyond just smoke testing, but that's probably a decent cost/reward tradeoff right now.  (The next step I might imagine is something hacky using testmark's script feature to cat some files and grep for key bits to sanity check... But that's asking relatively a lot, and wouldn't provide a ton more overall confidence.)

Nit: we should probably document that the url-prefix option can be be "/", and that that will DTRT for anyone about to publish as a website at the root of a subdomain.   The default to be the same as the local output path makes sense though, for when not publishing via a server.

PR #21, opened 2022-10-06
workspace stack will now end on finding a root workspace

TripleDogDare:
This seems like more correct behavior unless there's a reason that we don't want the stack to stop on a root workspace.


warpfork:
Yeah, I agree with the logic here.

(I think this is my old code that you're correcting here, and I think it was written before we had entirely figured out what this should act like, and so I vote "+2" that you're nailing down the correct behavior.)

warpfork (approved):
Yeah, I agree with the logic here.

(I think this is my old code that you're correcting here, and I think it was written before we had entirely figured out what this should act like, and so I vote "+2" that you're nailing down the correct behavior.)

warpfork:
... hm.  I might've expected this to change the behavior of `wf status`, slightly... but it doesn't.  It seems that command already doesn't report the home workspace path if it encounters some other root workspace, so it must have filtered out the home workspace somehow.

I wonder if there's code there that can be simplified now, with this change?

TripleDogDare:
There are a number of simplifications if we define a workspace set as a stack of workspaces ending with a root workspace.

PR #22, opened 2022-10-08
Cleanup workspace set

TripleDogDare:


cmd/warpforge/catalog.go
```diff
@@ -59,32 +59,32 @@ var catalogCmdDef = cli.Command{
...
 		{
 			Name:   "show",
-			Usage:  "Show the contents of a module",
+			Usage:  "Show the contents of a module in the root workspace catalog",
 			Action: cmdCatalogShow,
 		},
 		{
 			Name:   "bundle",
-			Usage:  "Bundle required catalog items into this project's catalog",
+			Usage:  "Bundle required catalog items into the current directory's workspace catalog. Will create a workspace and catalog in the current directory if none exists.",
```
warpfork, line 72: ... I hope this isn't what it does?  Creating a workspace, that is?  If it doesn't find one it shouldn't make one.

I'd hope it's accurate to say it does this:

```suggestion
			Usage:  "Bundles any catalog information used by any modules in the current workspace into this workspace's catalog.  Use this to make a workspace independent.",
```

pkg/workspace/catalog_test.go
```diff
@@ -100,10 +100,11 @@ func TestCatalogLookup(t *testing.T) {
 				},
 			}
 			var err error
-			wss, _, err := FindWorkspace(fsys, "", "home/user/")
+			ws, _, err := FindWorkspace(fsys, "", "home/user/")
```
warpfork, line 103: I notice all these consistency fixes and quietly appreciate them :)

pkg/workspace/workspace_set.go
```diff
@@ -6,46 +6,32 @@ import (
...
 
 // Opens a full WorkspaceSet
 // searches from searchPath up to basisPath for workspaces
-// root workspace will be the first workspace found that is marked as a root, or the home workspace if none exists
+// The root workspace will be the first workspace found that is marked as a root, or the home workspace if none exists
 // Errors:
 //
 //    - warpforge-error-workspace -- when the workspace directory fails to open
 //    - warpforge-error-searching-filesystem -- when an error occurs while searching for the workspace
 func OpenWorkspaceSet(fsys fs.FS, basisPath string, searchPath string) (WorkspaceSet, wfapi.Error) {
```
warpfork, line 23: Can we just nuke this function entirely now, then?  FindWorkspaceStack seems to already do everything.

In particular, FindWorkspaceStack is already doing the "include home workspace if no other root found" check.

TripleDogDare: I think I lean toward removing the `FindWorkspace*` functions all together in favor of dumping that into the `OpenWorkspace*` functions. The `Find` functions are calling `openWorkspace` anyway.

cmd/warpforge/catalog.go
TripleDogDare: Uhh well...
https://github.com/warpfork/warpforge/blob/4928d809fb1ed3cbe2e3e6a2d7f1e508a4c53c02/cmd/warpforge/catalog.go#L359-L362
Yeah I think that's what it does. For this PR I opted for describing _current_ behavior (AFAICT at least) over trying to make it the _desired_ behavior.

warpfork: Oh dear.  You're right.  And that's also definitely wrong, behaviorally.  Let's... fix that real quick, right after this >.>

pkg/workspace/workspace_set.go
warpfork: If it's just naming at that point... I think the `Find*` names make more sense than the `Open*` names, with what is being done in these functions.  There's not a lot of opening going on; there's no closing to be done afterwards either.

I'd tend to expect a function named `Open*` to be opening files (not just stat'ing their existence), and probably outright hanging on to open handles afterwards.

TripleDogDare: Perhaps the other way then. But `Open` still feels more correct than "Find" for the base `OpenWorkspace`, `OpenHomeWorkspace`, and `OpenRootWorkspace`. Perhaps `Get` would be fine. `New` feels incorrect because it fails when it would have to do a `Create` operation.

warpfork: Still think `Find` describes what's currently happening, which is: it returns a list of paths that look like workspaces.

When we do have workspace config that has some richness: then we'll end up having an `OpenWorkspace` operation that opens some files, reads and parses them, and perhaps does other additional thinking.  But that will be a distinct operation from finding the workspaces: we probably wouldn't even run it on every path returned by `FindWorkspace` in the course of many commands, I suspect.

TripleDogDare: It seems like there's a dichotomy between what happens internally "create a new workspace struct" and externally "there should be a file here and it is an error if there isn't"

pkg/workspace/workspace.go
```diff
@@ -38,30 +39,23 @@ func OpenWorkspace(fsys fs.FS, rootPath string) (*Workspace, wfapi.Error) {
...
+// from an outside perspective.
 func openWorkspace(fsys fs.FS, rootPath string) *Workspace {
 	rootPath = filepath.Clean(rootPath)
+	isHomeWorkspace := rootPath == homedir
 	return &Workspace{
 		fsys:            fsys,
 		rootPath:        rootPath,
-		isHomeWorkspace: rootPath == homedir,
+		isHomeWorkspace: isHomeWorkspace,
+		isRootWorkspace: checkIsRootWorkspace(fsys, rootPath) || isHomeWorkspace,
```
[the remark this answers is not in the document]
TripleDogDare: tl;dr; this is a filesystem operation that happens once on opening the workspace.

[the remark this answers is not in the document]
warpfork: Yeah, this seems fine.  I think we generally treat this scale of caching over filesystem as fine (it's not generally expected to be modified in normal operation; there's no significant secure implications; and there's not really any reasonable alternative given the general nature of multiuser filesystems).

warpfork (approved):

PR #23, opened 2022-10-11
HTML Replays & Download Links

ericevenchick:
This adds HTML output for replay plots and download links to release pages. The replay plots are indented, syntax highlighted (using [Chroma](https://github.com/alecthomas/chroma)). Catalog refs in the replay plots are also given links to the relevant release.

It has some issues currently: 

1. The replay link doesn't seem to be clickable (on Firefox at least)
2. Syntax highlighting doesn't follow the style background color.

### DIFF
--- a/pkg/cataloghtml/css.css
+++ b/pkg/cataloghtml/css.css
@@ -102,6 +102,15 @@ body > input {
 	min-height: 100%;
 }
 
+code {
+	overflow-x: auto;
+	white-space: pre-wrap;
+	white-space: -moz-pre-wrap;
+	white-space: -pre-wrap;
+	white-space: -o-pre-wrap;
+	word-wrap: break-word;
+}
+
 /* Structural classes */
 
 .header {
--- a/pkg/cataloghtml/css.css
+++ b/pkg/cataloghtml/css.css
@@ -103,12 +103,7 @@ body > input {
 }
 
 code {
-	overflow-x: auto;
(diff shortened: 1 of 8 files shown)


pkg/cataloghtml/css.css
```diff
@@ -102,6 +102,15 @@ body > input {
 	min-height: 100%;
 }
 
+code {
+	overflow-x: auto;
+	white-space: pre-wrap;
+	white-space: -moz-pre-wrap;
+	white-space: -pre-wrap;
+	white-space: -o-pre-wrap;
+	word-wrap: break-word;
```
Storken, line 111: Is it necessary to include browser-support for Mozilla back to 1999, Opera 4-7 and IE 5.5+? 😅 Or can we live with just `white-space: pre-wrap` for css-3, which is supported by all modern browsers? 

I guess I kinda like `white-space: pre-line` a bit more, but depends of course on what it will be used for. [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/white-space) has a really nice playground for testing the differences.

Storken: Ah shoot, this comment was meant for row 107-110

ericevenchick: Yup this was just some copypaste fail, thanks for the input. I just pre-wrap/pre-line should be good 👍

warpfork (approved):
Neat!!

PR #25, opened 2022-10-12
Update tracing constants and add more direct tracing around exec

TripleDogDare:
Some new standards for tracing bits and actually putting tracing around exec/git operations for testing feel.

TODO: I do need to actually _run_ this to see if it behaves okay.

### DIFF
--- a/cmd/warpforge/status.go
+++ b/cmd/warpforge/status.go
@@ -10,13 +10,19 @@ import (
 	"github.com/fatih/color"
 	"github.com/urfave/cli/v2"
 	"github.com/warpfork/warpforge/pkg/formulaexec"
+	"github.com/warpfork/warpforge/pkg/tracing"
 	"github.com/warpfork/warpforge/wfapi"
+	"go.opentelemetry.io/otel/trace"
 )
 
 var statusCmdDef = cli.Command{
-	Name:    "status",
-	Usage:   "Get status of workspaces and installation",
-	Action:  cmdStatus,
+	Name:  "status",
+	Usage: "Get status of workspaces and installation",
+	Action: chainCmdMiddleware(cmdStatus,
+		cmdMiddlewareLogging,
+		cmdMiddlewareTracingConfig,
+		cmdMiddlewareTracingSpan,
+	),
 	Aliases: []string{"info"},
 }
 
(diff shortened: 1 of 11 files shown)


cmd/warpforge/status.go
```diff
@@ -10,7 +10,9 @@ import (
 	"github.com/fatih/color"
 	"github.com/urfave/cli/v2"
 	"github.com/warpfork/warpforge/pkg/formulaexec"
+	"github.com/warpfork/warpforge/pkg/tracing"
 	"github.com/warpfork/warpforge/wfapi"
+	"go.opentelemetry.io/otel/trace"
```
warpfork, line 15: Nit: I'm beginning to gain a desire that we enforce goimports in CI.  (e.g. imports from this same module go in a separate block.)

pkg/tracing/tracing.go
```diff
@@ -52,16 +54,36 @@ func Start(ctx context.Context, spanName string, opts ...trace.SpanStartOption)
...
+
+// Printable returns the input with only printable characters as defined by unicode.IsGraphic
+func Printable(s string) string {
+	return strings.Map(func(r rune) rune {
+		if unicode.IsGraphic(r) {
+			return r
+		}
+		return -1
+	}, s)
+}
```
warpfork, line 89: I wonder if perhaps you might be looking for https://pkg.go.dev/strconv#QuoteToGraphic here?

Although also I should note that the only place I saw this being used was on step names, and those are not meant to permit totally wild nongraphical characters either.  (Enforcement, admittedly, probably not-yet-implemented.)  So this might be slightly overkill / adding enforcement against wild stepnames might be a better investment.

warpfork:
Tentative lgtm.  It textually looks fine to me.  I'll trust your familiarity with the tracing on those outcomes being improved!

pkg/tracing/tracing.go
TripleDogDare: Hmm, I don't really want the surrounding quotes though. I do generally think we are going to want some kind of user-input validation/protection but I'm not sure what that looks like exactly. I was thinking that I might run a similar filter on the logging output in the hopes of fixing the tty bug.

cmd/warpforge/status.go
TripleDogDare: Which `goimports` version/invocation are you using? Mine does not do this. I think it respects line separation if you do it manually but it won't enforce it.

warpfork: :see_no_evil: I'm not using it either; I thought that was a feature, though.  Maybe I don't mean `goimports` and I'm thinking of some other tool...

TripleDogDare:
Updated all the commands to enable tracing/logging middleware.

PR #26, opened 2022-10-13
Clean up bundle behavior

TripleDogDare:
Make catalog bundle work on the most local workspace without creating a new workspace.

### DIFF
--- a/pkg/workspace/workspace.go
+++ b/pkg/workspace/workspace.go
@@ -283,15 +283,15 @@ func (ws *Workspace) HasCatalog(name string) (bool, wfapi.Error) {
 	return true, nil
 }
 
-// createCatalog creates a new catalog
-// createCatalog does not open the new catalog.
+// CreateCatalog creates a new catalog.
+// CreateCatalog only creates the catalog and does not open it.
 //
 // Errors:
 //
 //    - warpforge-error-io -- when reading or writing the catalog directory fails
-//    - warpforge-error-catalog-invalid -- when the catalog already exists
+//    - warpforge-error-file-exists -- when the catalog already exists
 //    - warpforge-error-catalog-name -- when the catalog name is invalid
-func (ws *Workspace) createCatalog(name string, existsOk bool) wfapi.Error {
+func (ws *Workspace) CreateCatalog(name string) wfapi.Error {
 	path, err := ws.CatalogPath(name)
 	if err != nil {
 		return err
@@ -304,10 +304,7 @@ func (ws *Workspace) createCatalog(name string, existsOk bool) wfapi.Error {
 		return err
 	}
 	if exists {
-		if !existsOk {
-			return wfapi.ErrorCatalogInvalid(path, "catalog already exists")
-		}
-		return nil
+		return wfapi.ErrorFileExists(path)
 	}
 
 	errRaw := os.MkdirAll(path, 0755)
@@ -326,22 +323,17 @@ func (ws *Workspace) createCatalog(name string, existsOk bool) wfapi.Error {
 //  - warpforge-error-catalog-name -- when the catalog name is invalid
 //  - warpforge-error-catalog-invalid -- when opened catalog has invalid data
 func (ws *Workspace) CreateOrOpenCatalog(name string) (Catalog, wfapi.Error) {
-	if err := ws.createCatalog(name, true); err != nil {
-		return Catalog{}, err
+	err := ws.CreateCatalog(name)
+	if err != nil {
+		switch err.(*wfapi.ErrorVal).Code() {
+		case "warpforge-error-file-exists":
+			return ws.OpenCatalog(name)
+		default:
+			// Error Codes -= warpforge-error-file-exists
+			return Catalog{}, err
+		}
 	}
 	return ws.OpenCatalog(name)
-}
-
-// Create a new catalog.
-// This only creates the catalog and does not open it.
-//
-// Errors:
-//
-//     - warpforge-error-io -- when reading or writing the catalog directory fails
-//     - warpforge-error-catalog-invalid -- when the catalog already exists
-//  - warpforge-error-catalog-name -- when the catalog name is invalid
-func (ws *Workspace) CreateCatalog(name string) wfapi.Error {
-	return ws.createCatalog(name, false)
 }
 
 // Get a catalog replay from a workspace, doing lookup by CatalogRef.
--- a/pkg/workspace/workspace.go
+++ b/pkg/workspace/workspace.go
@@ -171,7 +171,7 @@ func (ws *Workspace) CatalogPath(name string) (string, wfapi.Error) {
 //
 // Errors:
 //
-//    - warpforge-error-catalog-invalid -- when opened catalog has invalid data or an invalid name
+//    - warpforge-error-catalog-invalid -- when opened catalog has invalid data
 //    - warpforge-error-io -- when IO error occurs during opening of catalog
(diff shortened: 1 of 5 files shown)


pkg/workspace/workspace.go
```diff
@@ -195,8 +195,7 @@ func (ws *Workspace) ListCatalogs() ([]*string, wfapi.Error) {
 }
 
 // Get a catalog ware from a workspace, doing lookup by CatalogRef.
-// This will first check all catalogs within the "catalogs" subdirectory, if it exists
-// then, it will check the "catalog" subdirectory, if it exists
+// This will first search catalogs within the "catalogs" subdirectory and then search the "catalog" subdirectory
```
warpfork, line 198: Hm.  Okay, feel free to deflect this as "not in this PR", but...

Now that I'm looking at it, this function below is doing kind of a silly thing.

The behavior I think would be most excellent would be to check the nearest catalog first, then check any other catalogs seen on the way to the root and make sure they match, and then ultimately do the range over root catalogs and make sure it exists in exactly one of them and still matches.

(Or more realistically, that should probably be a few different functions, or, just one function with a bunch of flags to control how much work it invests in active sanity checks.)

warpfork:
This looks probably correct.

How much of a stretch is it to set up a test that covers this behavior path?  Is there enough infrastructure that it would be easy to add, or do we have work to do there?

pkg/workspace/workspace.go
TripleDogDare: >check the nearest catalog first, then check any other catalogs seen on the way to the root and make sure they match

This is something that could work for the `WorkspaceSet` implementation. For a single workspace it might not make sense. However you could have it only check `catalogs` directory if it's a root workspace if that's a dichotomy we want to enforce.

TripleDogDare:
>How much of a stretch is it to set up a test that covers this behavior path? Is there enough infrastructure that it would be easy to add, or do we have work to do there?

We don't have a lot of testing for the `main` package and CLI yet outside the testmark stuff which seems insufficient for the kind of testing we want to do here. Honestly I would rework a lot of these commands to be `pkg` level APIs that get configured and called by the CLI. Although I don't know if testing there would obviate the need for more CLI testing.

TripleDogDare:
After further attempts at writing tests, I'm going to say needs a bit of infrastructure and I should plop that into it's own thing.

It _does_ seem to work when bundling the quickstart.
```
 ~/repos/warpforge/wf (change-bundle-behavior) $ rm -rf .warpforge/catalog/* && warpforge catalog bundle && tree .warpforge/catalog/
bundled "min.warpforge.io/alpinelinux/rootfs:v3.15.4:amd64"
.warpforge/catalog/
└── min.warpforge.io
    └── alpinelinux
        └── rootfs
            ├── _mirrors.json
            ├── _module.json
            └── _releases
                └── v3.15.4.json

4 directories, 3 files
```
However it errors if the catalog directory is missing.
```
 ~/repos/warpforge/wf (change-bundle-behavior) 1 $ rm -rf .warpforge/catalog && warpforge catalog bundle
bundled "min.warpforge.io/alpinelinux/rootfs:v3.15.4:amd64"
error: failed to open catalog: invalid catalog file "home/cjb/repos/warpforge/wf/.warpforge/catalog": catalog does not exist
```

warpfork:
I think this code should probably feel free to go ahead and make the `.warpforge/catalog` dir if it doesn't exist.  (Making a new workspace directory -> bad, that's a big behavioral action.  Maintaining the internals of an existing workspace directory quietly by default -> fine and usually good.)

TripleDogDare:
I'd like an alternative name for `CreateOrOpenCatalog`. Currently it does a `CreateCatalog` ignoring any `already exists` errors before an `OpenCatalog`. The name feels incorrect but I also don't want to yak shave too much.

pkg/workspace/workspace.go
```diff
@@ -283,16 +283,15 @@ func (ws *Workspace) HasCatalog(name string) (bool, wfapi.Error) {
...
+// CreateCatalog creates a new catalog.
+// CreateCatalog only creates the catalog and does not open it.
 //
 // Errors:
 //
-//     - warpforge-error-io -- when reading or writing the catalog directory fails
-//     - warpforge-error-catalog-invalid -- when the catalog already exists
-//     - warpforge-error-catalog-name -- when the catalog name is invalid
+//    - warpforge-error-io -- when reading or writing the catalog directory fails
+//    - warpforge-error-file-exists -- when the catalog already exists
```
warpfork, line 292: I'd probably rather we have "warpforge-error-already-exists" and leave the file detail out of the headline?

TripleDogDare: Should it still record path or any other details?

TripleDogDare:
```
		Message:    "already exists"
```
Seems a bit sparse.

warpfork: I think putting a path or mention files in the message text or details map is fine.  More info is better in those!  I'm only objecting to mentioning files in the code itself.

TripleDogDare: So the constructor implicitly means files but the error code itself doesn't. I.E. The code is reusable for multiple constructors?

TripleDogDare: We could then change `warpforge-error-catalog-item-already-exists` to return a code of `warpforge-error-already-exists`. I'm not really sure if that's _favorable_ but would reduce raw quantities of error codes.

TripleDogDare: Also `warpforge-error-already-exists` seems excessive compared to `warpforge-error-exists`.

warpfork: - the constructors could definitely be re-steered.  I think it's fine to have more than one constructor that produces an error with the same code, but with different parameters because it's describing a slightly different origin story for the error.
- It's hard to say for sure, because we're still figuring out what good heuristics for error codes are, but I think I'm :+1: on `warpforge-error-catalog-item-already-exists` -> `warpforge-error-already-exists`, yes!  (In both cases, the remedy action would not be to retry, but would involve changing a parameter in some direction to avoid a collision... so, for these things to have the same error code seems logically valid.  And also because they tend not to emerge up along the same code paths, it is unlikely to be confusing.  It's passing every heuristic I can think of at the moment!)
- :-1: on stripping the word "already" out.  Maybe this is personal taste, but I think that _does_ provide useful information to a human being.  Sure, if words had no inherent meaning and we gziped everything, `already-exists` and `exists` could be collapsed to the same shortcode in practice.  But that's not how humans process things.

wfapi/error.go
```diff
@@ -382,3 +386,18 @@ func ErrorCatalogName(name string, reason string) Error {
...
+//
+// Errors:
+//
+//    - warpforge-error-already-exists --
+func ErrorFileAlreadyExists(path string) Error {
+	return &ErrorVal{
+		CodeString: errCodeAlreadyExists,
+		Message:    fmt.Sprintf("file already exists at path: %q", path),
+		Details: [][2]string{
+			{"path", path},
```
warpfork, line 400: fwiw, this is all going to get better when we switch over to using the [go-serum](https://github.com/serum-errors/go-serum) package instead of our hand-rolled things here.  It allows attaching causes easily with either `%w` in `serum.Errorf` constructor, or `serum.WithCause` for the fuller `serum.Error` constructor.  So this could probably disappear into a puff of simply attaching the file IO error as a cause.

warpfork (approved):
I'm good with all this

PR #27, opened 2022-10-14
Update catalog name handling to not use a pointer

TripleDogDare:
Also adds catalog name validation.


pkg/workspace/catalog.go
```diff
@@ -18,8 +19,11 @@ import (
 
 const (
 	magicModuleFilename = "_module.json"
+	CatalogNameFormat   = `^([A-Za-z0-9][-A-Za-z0-9_.]{0,62})?$`
```
warpfork, line 22: Seems a little aggro.  This doesn't bear a ton of load, either, because it's really only a local name.  It doesn't get serialized and shipped around in other key persistent data structures.

That said: it's easy to loosen this up later, than to lock it down later, so... :+1:

warpfork: Hang on, what's the `?` for?

Also do you need the parens for a capture grouping?  I think not, because you're only using it to check for a match overall, not to extract things, yes?

pkg/workspace/workspace.go
```diff
@@ -134,20 +135,30 @@ func (ws *Workspace) CatalogBasePath() string {
...
+// defaultCatalogPath returns the path to the default catalog belonging to this workspace.
+func (ws *Workspace) defaultCatalogPath() string {
+	return filepath.Join(ws.rootPath, ".warpforge", "catalog")
+}
+
+// Returns the catalog path for catalog with a given name within a workspace.
+// Guards against filepath modifying names by considering any path which
+// would be modified by filepath.Clean to be considered invalid.
+// Guards against catalog nesting by considering any name with filepath separators to be invalid.
+//
```
warpfork, line 147: Maybe worth a comment here about empty string returning the single catalog in a non-root workspace

pkg/workspace/workspace.go
```diff
@@ -134,20 +135,30 @@ func (ws *Workspace) CatalogBasePath() string {
...
+// Errors:
+//
+//    - warpforge-error-catalog-invalid -- when catalog name is invalid
+func (ws *Workspace) CatalogPath(name string) (string, wfapi.Error) {
+	if name == "" {
+		return ws.defaultCatalogPath(), nil
 	}
```
warpfork, line 154: This should return an error if we're in a root workspace, come to think of it.

pkg/workspace/workspace.go
```diff
@@ -134,20 +135,30 @@ func (ws *Workspace) CatalogBasePath() string {
...
+}
+
+// Returns the catalog path for catalog with a given name within a workspace.
+// Guards against filepath modifying names by considering any path which
+// would be modified by filepath.Clean to be considered invalid.
+// Guards against catalog nesting by considering any name with filepath separators to be invalid.
+//
+// Errors:
+//
+//    - warpforge-error-catalog-invalid -- when catalog name is invalid
```
warpfork, line 150: Hm.  I think this should get a new error code.

Reason: who to blame is distinct.  Elsewhere, `warpforge-error-catalog-invalid` is used when we encounter invalid data _in_ the catalog already.  Here, the blame isn't in the catalog: it's in the parameter handed into this function.

warpfork (changes requested):
A few really minor change requests, and one medium one (new error code plz).

Otherwise looks great to me!

pkg/workspace/catalog.go
TripleDogDare: The `()?` allows the validator to accept an empty string. We _could_ remove it because we are handling the empty string explicitly. But it's not a huge deal either way.

pkg/workspace/workspace.go
TripleDogDare: Yeah I thought about that aaaand ignored it because I didn't want to deal with it.

pkg/workspace/workspace.go
```diff
@@ -156,74 +173,81 @@ func (ws *Workspace) CatalogPath(name *string) string {
...
 	for _, c := range cats {
 		cat, err := ws.OpenCatalog(c)
 		if err != nil {
-			return nil, nil, err
+			switch err.(*wfapi.ErrorVal).Code() {
+			case "warpforge-error-catalog-name":
+				err := wfapi.ErrorInternal("referenced an invalid catalog", err)
+				return nil, nil, err
+			default:
+				// Error Codes -= warpforge-error-catalog-name
```
TripleDogDare, line 248: @warpfork This might be a controversial decision. Not sure if you'd prefer it stay as a `warpforge-error-catalog-name` even though in this instance it really shouldn't happen because the names are already validated by `ListCatalogs`.

And I do it again [here](https://github.com/warpfork/warpforge/pull/27/files#diff-92c06c458c8332afdcac283317a34cb1302e1b722216cd26d10056c9c2bfb4c0R341-R349)

TripleDogDare: This was updated to panic instead of returning an internal error. It still removes the error-catalog-name from the error output.

warpfork: SGTM.

PR #28, opened 2022-10-20
Disable wiring stdin to the container unless interactive mode is enabled

ericevenchick:
Previously, `os.Stdin` was always wired to the container unless `/dev/tty` didn't exist. This was due to an oversight when enabling the interactive mode for `ferk`.

This disables wiring stdin to the container unless interactive mode is enabled. It appears to fix the need to run `reset` to fix your shell after running tests 🥳

### DIFF
--- a/pkg/formulaexec/formula_exec.go
+++ b/pkg/formulaexec/formula_exec.go
@@ -47,10 +47,11 @@ type RioOutput struct {
 }
 
 type runConfig struct {
-	spec    specs.Spec
-	runPath string // path used to store temporary files used for formula run
-	wsPath  string // path to of the workspace to run in
-	binPath string // path containing required binaries to run (rio, runc)
+	spec        specs.Spec
+	runPath     string // path used to store temporary files used for formula run
+	wsPath      string // path to of the workspace to run in
+	binPath     string // path containing required binaries to run (rio, runc)
+	interactive bool   // flag to determine if stdin should be wired to containier for interactivity
 }
 
 // base directory for warpforge related files within the container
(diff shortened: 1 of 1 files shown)


pkg/formulaexec/formula_exec.go
```diff
@@ -504,7 +506,9 @@ func invokeRunc(ctx context.Context, config runConfig, logWriter io.Writer) (str
 		fmt.Sprintf("warpforge-%d", time.Now().UTC().UnixNano()), // container id
 	)
 
-	if config.spec.Process.Terminal {
+	// if the config has terminal enabled, and interactivity is requested,
+	// wire stdin to the contaniner
+	if config.spec.Process.Terminal && config.interactive {
 		cmd.Stdin = os.Stdin
```
TripleDogDare, line 512: Should we be passing an `io.Reader` instead of using `os.Stdin` directly?

warpfork (approved):
oh thank goodness

pkg/formulaexec/formula_exec.go
warpfork: Absolutely not.  The `os.Stdin` handle "happens to be" also a device file which is imbued with all the tty wondersauces.  Passing a golang level buffer around would cause drastically different behaviors.

(... usually, of course.  If you were to pipe things into the warpforge process, then of course stdin already isn't a tty.  IIUC, that's part of what the whole `config.spec.Process.Terminal` reference is handling.)

warpfork:
This whole surrounding area is a bit due for refactoring.  :/  Trying to rediscover what this code does without the comments would be... difficult.  There's several pieces of spooky-action-at-a-distance already cropping up.

E.g., where the interactive flag gets set... well, that's half of the effective control for what happens.  The other half is where `config.spec.Process.Terminal` is consulted.  Which comes from... even further yet away.  Even the first two points are 400+ lines apart; and embedded in 300+ line functions.  One has to read a very large amount of code to figure out what the real source of behaviors is.

I don't think this can be improved in a small PR, so we should probably merge this (rather important) fix, and move on iteratively.  But in reviewing this area, it's not surprising a bug like this was able to leak in here in the first place.  We're gonna have to do better code organization than this going forward.

ericevenchick:
Strong agree on refactoring. This was the very first code of warpforge and it's grown to support various things. 

A refactor here probably means defining the API better then rewriting it to be less... crazy. At the same time, we should consider what it would look like to have other (non runc) executors and build an API to suit.

But, that's not a today problem so I'll click the big merge button now.

PR #30, opened 2022-10-26
Add health check command

TripleDogDare:
This is the general gist of the health check command so far.
I have some plans for extending it some more _but_ it might be better to merge this in as it is already larger than I am generally comfortable with and includes a lot of refactoring.


warpfork (approved):
I skimmed at high speed but this LGTM!  It's progress towards making a bunch of the CLI stuff less tangled :+1: 

Health check component now appears to work, as of that last push, too.

PR #34, opened 2022-11-01
Release page mobile

Storken:
## Description
We currently have some issues showing a few pages on mobile, some have a lot of content which breaks the view like this:
image

Here are some changes that fixes this issue and makes for a much better mobile experience.


warpfork:
I think h4 seems a little aggro, semantically, for some of this content?

But otherwise the rendered output seems improved so I shall merge it :)

PR #36, opened 2022-11-09
Refactor html templates

Storken:
## Description
Refactor our css-files to follow the [BEM methodology](https://getbem.com/introduction/) to prepare for more css, which can be tricky now that we are starting to have a lot of it.

Also introduce a index.css barrel which will import all the css files we use, so that we only need to reference one css-file in the html-templates. (easier to maintain)


Storken:
Do it!🥳

warpfork:
(Whoops!  I guess my github event filtering systems _still_ need improvement!)

PR #39, opened 2022-11-28
fix up cli tests

TripleDogDare:
Huzzah! :partying_face: 

>With the new additions to testmark, we have a bunch of tests to fix that
weren't actually executing before. This gets us back into a relatively
good state where all the tests pass.

>This change also makes significant changes to how the offline "/net/"
flag works. I.E. Any non-executable children of the parent DirEnt are
copied to the "/net/" DirEnt before execution. This allows a parent with
both online and offline tests to execute and properly pass along
required filesystems, although the parent script/sequence may execute
twice. However, this is _not required_ and a parent of a "/net/"
directory does not need to be executable or have any other children if
the "/net/" DirEnt has executable children. The children copied to the
"/net/" DirEnt must be unique. Any duplicates will be an error.

### DIFF
--- a/cmd/warpforge/main_test.go
+++ b/cmd/warpforge/main_test.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"fmt"
 	"io"
 	"os"
 	"path/filepath"
@@ -29,6 +28,25 @@ func getTestWorkspaceStack(t *testing.T) []*workspace.Workspace {
 	return wss
 }
 
+type tagset map[string]struct{}
+
+func newTagSet(tags ...string) tagset {
+	result := tagset(make(map[string]struct{}))
+	for _, s := range tags {
+		s = strings.TrimSpace(s)
+		if s == "" {
+			continue
+		}
+		result[s] = struct{}{}
+	}
+	return result
+}
+
+func (t tagset) has(tag string) bool {
+	_, ok := t[tag]
+	return ok
+}
+
 func testFile(t *testing.T, fileName string, workDir *string) {
 	doc, err := testmark.ReadFile(fileName)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


TripleDogDare:
> "bash: line 1: warpforge: command not found"

:crab:

cmd/warpforge/main_test.go
```diff
@@ -48,19 +67,30 @@ func testFile(t *testing.T, fileName string, workDir *string) {
...
 	for _, dir := range doc.DirEnt.ChildrenList {
 		testName := dir.Name
+		tags := newTagSet()
 		testDir := dir
-		if _, hasNetTests := dir.Children["net"]; hasNetTests {
-			if *testutil.FlagOffline {
-				t.Log("skipping test", dir.Name, "due to offline flag")
-				continue
+		for _, x := range dir.ChildrenList {
+			//iterating on children just to find out if it has tags sucks but it's the only way
```
warpfork, line 76: seems a bit much to say this sucks :P

"iterating on data necessary in order to respond to it", oh no :P

cmd/warpforge/main_test.go
```diff
@@ -28,6 +28,25 @@ func getTestWorkspaceStack(t *testing.T) []*workspace.Workspace {
 	return wss
 }
 
+type tagset map[string]struct{}
```
warpfork, line 31: I like this whole bit :+1:   A good terse little abstraction for this feels nice and makes it clearer what's going on.

examples/500-cli/cli.md
```diff
@@ -364,21 +364,21 @@ there's only one record in this map.
...
-[testmark]:# (catalog/sequence)
+[testmark]:# (catalog/tags=net/sequence)
 ```
 warpforge catalog init my-catalog
 ```
 
 ### List Catalogs
 
-[testmark]:# (catalog/then-ls/sequence)
+[testmark]:# (catalog/tags=net/then-ls/sequence)
```
warpfork, line 381: I like having this "tags=" thing here for legibility.  We should've done this the first time around, and I'm happy to see it now.

examples/500-cli/cli.md
```diff
@@ -428,7 +427,7 @@ cat .warpforge/catalogs/my-catalog/warpsys.org/busybox/_mirrors.json
 	"catalogmirrors.v1": {
 		"byWare": {
 			"tar:4z9DCTxoKkStqXQRwtf9nimpfQQ36dbndDsAPCQgECfbXt3edanUrsVKCjE9TkX2v9": [
-				"file://.warpforge/warehouse/4z9/DCT/4z9DCTxoKkStqXQRwtf9nimpfQQ36dbndDsAPCQgECfbXt3edanUrsVKCjE9TkX2v9"
```
warpfork, line 431: Okay, so, I know we're plowing through some things that weren't getting exercised, so maybe there's been rot here -- but ideally, this stuff using local "file://" _should_ work, and we have vendored that file in the git repo for this purpose, and I think that if that works and the file is present, then hopefully these tests don't actually need network?

warpfork (approved):
Seems to be headed in a good direction.

A couple questions about if tests actually need to be networked or if we have more fixing to do there.

cmd/warpforge/main_test.go
TripleDogDare: Well we went from O(1) to O(n) on that operation so my jimmies are rustled.

examples/500-cli/cli.md
TripleDogDare: hmm possibly. Not sure when/why/how that got switched over.

TripleDogDare: Ah no that's entirely just the output of the command.
https://github.com/warptools/warpforge/pull/39/files/4f504b1c0a2a1a98b2e5886d01d4678acf6c536f#diff-8d8e5d1c040d3a0eab640b364644a15d331f15d74fd1d3ca0bd5115ac84f8623L399

We don't have a way to specify the location of the file without going into shell mode.

TripleDogDare: We could feasibly _actually_ copy the warehouse into the workspace during the ExecFn instead of just setting an env var for it.

examples/500-cli/cli.md
TripleDogDare: Yeah. It occurs to me that for testmark parsing purposes the tags must _always_ be in the same order. Might annoy some people but it's probably fine. Makes it easy to sed replace later at least.

cmd/warpforge/main_test.go
TripleDogDare: I cleaned this up a bit. Still O(n) but at least it's not using labels.

examples/500-cli/cli.md
ericevenchick: +1 to "We should've done this the first time around" :+1:

cmd/warpforge/main_test.go
ericevenchick: Same, though a nit... we might want to use this elsewhere as well. Maybe worth extracting the tags bits into a util package?

Could happen later though.

ericevenchick (approved):
Generally LGTM, though we may wish to extract the tag stuff into testutil for use elsewhere in the future.

Also, all tests green on my machine :smiley:

cmd/warpforge/main_test.go
warpfork: Vote for later.  I appreciate that this is a very small hunk of code, and that it doesn't have to take itself too seriously because it has exactly one user that's very local

examples/500-cli/cli.md
warpfork: (Maybe it should be WARPFORGE_FILE_WAREHOUSE_PREFIXPATH or something, on second thought, because maybe we wanted to explicitly have sone mirrors info file here?  I don't remember how explicit we were trying to be in this particular example / maybe worth briefly revisiting the intent.)

ericevenchick: We never did the `EXTRA_WAREHOUSE` thing. Getting this to work was generally a pain with paths, so that might be the way to go.

TripleDogDare: What's the difference between those and `WARPFORGE_WAREHOUSE`?

PR #40, opened 2022-11-29
Ferk panic

TripleDogDare:
Relatively minor fix. I will however take a moment to complain about ipld error output.
>execution failed: error from ExecFn is "warpforge-error-plot-invalid: plot file ./plot.wf not parsed: caused by: warpforge-error-serialization: serialization error: unable to deserialize plot: bindnode TODO: len mismatch"

Particularly that last bit: `bindnode TODO: len mismatch` is where the actual "which part of this json is bad" error should be sitting. Playing around with the ipld library, I got it to emit `could not assign string \"pipe:rootfs\": bindnode TODO: len mismatch` which is significantly more useful (oh look I forgot a colon). I'm not sure if that kind of thing is generalizable in any way.

Oh and this is dependent on #39. <- merged and rebased.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,7 +1,5 @@
 module github.com/warptools/warpforge
 
-//replace github.com/ipld/go-ipld-prime => /home/eric/git/go-ipld-prime
-
 go 1.18
 
 require (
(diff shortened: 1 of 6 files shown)


go.mod
```diff
@@ -1,6 +1,7 @@
 module github.com/warptools/warpforge
 
 //replace github.com/ipld/go-ipld-prime => /home/eric/git/go-ipld-prime
+//replace github.com/ipld/go-ipld-prime => ../go-ipld-prime
```
warpfork, line 4: We should probably remove both this and the old one, which shouldn't have been here either :)

warpfork:
LGTM

TripleDogDare:
Removed ipld replace comments from go.mod

PR #41, opened 2022-12-01
Ci socket

TripleDogDare:
Phase 1 of CI for warpforge is to set up a service with watch that can emit data. No client implementation. No RPCs or protocols. Just shoving an int down a pipe.

Questions we answer here are:
1. CLI API for server setup
2. Configuring and starting server
3. A single server type implementation: I.E. Currently a unix socket.
4. Server setup, shutdown, naming conflicts

Punts:
   - how to handling linux socket name length limits? Will likely need further iterations.
   - where to put sockets? No. We will need to iterate on this.
   - how to clean up sockets?
     - Mostly don't. They either go inside a workspace directory where they will never really be seen or in /tmp where they can be cleaned up _eventually_.
     - Use deterministic (ish?) naming so that we don't generate absurd amounts of files for most users and reuse sockets that aren't active.
     - Optimistic use of an interrupt handler.


warpfork:
Okay, I dunno if today is the day to pick this bone, but: in the future, I'd really like to get pretty much everything out of the `cmd/*` packages that isn't essentially args parsing.  And I don't really like `internal` packages either.

(Opinions may vary on "internal" packages, but for me: The number of times I've said "ah, having an internal package is so freeing" vs the number of times I've said "good lord this would be easy if someone hadn't sprinkled an internal package in here" is _extremely_ onesided and it's not in good favor of the internal packages.)

I'd like to have a package somewhere -- under `pkg/*`, I guess -- that contains most of the "watch" feature.  Optionally, we could start splitting up the detection loop stuff and this socket stuff already.  Anything still left under the `cmd/*` package should figure out how to populate config structs for those two things (and wire the signal channels), and then hand off.

I see this is logically already almost the case, since the `cmdWatch` func is pretty much filling in the `watchCfg` struct and then calling a `run` func on it -- great!  Now let's just make the package structure enforce that too, and get a less cluttered `cmd/*`.

cmd/warpforge/watch.go
```diff
@@ -23,20 +32,132 @@ var watchCmdDef = cli.Command{
 	Action: util.ChainCmdMiddleware(cmdWatch,
 		util.CmdMiddlewareLogging,
 		util.CmdMiddlewareTracingConfig,
+		util.CmdMiddlewareCancelOnInterrupt,
 	),
+	Flags: []cli.Flag{
+		&cli.BoolFlag{
+			Name:  "socket",
+			Usage: "Experimental flag for getting execution status of the watched plot externally via unix socket",
+		},
```
warpfork, line 41: I'd almost argue we don't need a flag at all, because why would we ever not enable this?

But at least let's have it default to "on".

warpfork: (The only reason I can think of to not enable this if a user gets stuck in a horrible situation where they absolute cannot create socket files, and they want to try to use the rest of the "watch" feature anyway.  Perhaps possible.  But seems rare.)

cmd/warpforge/watch.go
```diff
@@ -23,20 +32,132 @@ var watchCmdDef = cli.Command{
...
+	log.Info("", "listener closed")
+	return <-result
+}
+
+func isSocket(m fs.FileMode) bool {
+	return m&fs.ModeSocket != 0
+}
+
+// rmUnixSocket will perform a racy, lockless "liveness" check on the socket before removing it.
+func (c *watchCfg) rmUnixSocket(path string) error {
```
warpfork, line 122: Needs more description of expected outcomes.

- If a socket is found, but someone's alive and answering on it already -> what happens?
- If a socket is found, but someone's _not_ answering it (explicit error) -> what happens?
- If a socket is found, but someone's not answering it in a timely fashion -> what happens?

And, from the other side, because this is a coordination problem scene:

- If another process _is_ alive, but failed to answer in a timely fashion, and we removed its socket -> what happens to it?  Does it notice?  What does it do when it does?

warpfork: We talked about file locking in chat earlier, so I'm curious -- is this the function you were thinking of wrapping it around?

Would it be able to make things significantly better?

(I think we may end up being okay merging something that's a first pass on this and has some admitted race conditions, as long we have it reasonably encapsulated so we can revisit it later.  Part of the reason I'm okay with this is because I actually don't know how solvable the TOCTOUs here _are_, in the limit.  But if you did end up with more notes on where exactly the future work could go, they might be worth attaching.)

cmd/warpforge/watch.go
```diff
@@ -23,20 +32,132 @@ var watchCmdDef = cli.Command{
...
+}
+
+func (c *watchCfg) serveLoop(ctx context.Context, l net.Listener) error {
+	log := logging.Ctx(ctx)
+	for {
+		conn, err := l.Accept() // blocks, doesn't accept a context.
+		if err != nil {
+			return err
+		}
+		go func() {
```
warpfork, line 74: I think it may be a good idea to habituate having some small struct for a task before hitting the `go` keyword.

This most definitely works, but I worry that it rapidly becomes harder to reason about, name the thing, and so forth.

Some functions are _so_ small that an anonymous func handed to the `go` keyword without further aplomb is fine (the `go CancelOnInterrupt` is an example I'd put on this side), but things that are bigger, or that we spawn many of, seem worth giving a bit heavier treatment.

In this particular case, I think there's a whole "actor" worth of code here and a struct to organize it will help.

We don't necessarily have to do it in this pass, but even if we choose to punt in this PR, I strongly suspect we'll extract this and introduce such structures in a second pass.

warpfork (commented):
Many comments :)

cmd/warpforge/watch.go
warpfork: (I might go play with supervisor frameworks for a bit today and come back with more specific opinions.  I won't block us on that, though.)

cmd/warpforge/watch.go
ericevenchick: Agree, should probably be on by default.

cmd/warpforge/watch.go
```diff
@@ -23,20 +32,132 @@ var watchCmdDef = cli.Command{
...
+	conn.Close()
+	return nil
+}
 
+func generateSocketPath(path string) string {
+	// This socket path generation is dumb. and also one of the simplest thing to do right now
+	sockPath := fmt.Sprintf("/tmp/warpforge-%s", url.PathEscape(path))
+	if len(sockPath) > 108 {
+		// There is a 108 char limit on socket paths. This is also dumb, but not our fault.
+		// Hopefully you don't run into this as a problem
```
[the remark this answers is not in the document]
ericevenchick: We also discussed hashing the path? Not sure if that is viable as a solution.

ericevenchick (approved):

cmd/warpforge/watch.go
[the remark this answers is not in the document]
TripleDogDare: The two major hurdles to that are 1) hashes can be really long. 2) hashes are opaque to users.
It's certainly an option to keep in mind going forward though.

cmd/warpforge/watch.go
```diff
@@ -64,14 +185,34 @@ func cmdWatch(c *cli.Context) error {
...
+		}
+		go func() {
+			ctx, cancel := context.WithCancel(ctx)
+			defer cancel()
+			if err := c.serve(ctx, sockPath); err != nil {
+				log.Info("", "socket server closed: %s", err.Error())
+			}
+		}()
+		log.Info("", "serving to %q\n", sockPath)
+		runtime.Gosched()
```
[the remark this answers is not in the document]
TripleDogDare: hmm... code wasn't long enough? :shrug:

cmd/warpforge/watch.go
```diff
@@ -111,3 +256,24 @@ func cmdWatch(c *cli.Context) error {
...
+				DisableMemoization: c.Bool("force"),
+			},
+		},
+	}
+	err := cfg.run(c.Context)
+	if errors.Is(err, context.Canceled) {
+		return nil
+	}
+	return err
+}
```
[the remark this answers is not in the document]
TripleDogDare: Yeah I've been meaning to get around to that... there's a bunch of dragons in there though. Might as well get started.

cmd/warpforge/watch.go
```diff
@@ -23,20 +32,132 @@ var watchCmdDef = cli.Command{
...
+//serve will create and listen to a unix socket on the given socket path
+func (c *watchCfg) serve(ctx context.Context, sockPath string) error {
+	log := logging.Ctx(ctx)
+	cfg := net.ListenConfig{}
+	l, err := cfg.Listen(ctx, "unix", sockPath)
+	if err != nil {
+		return err
+	}
+	// unix socks: they have permissions systems.
+	os.Chmod(sockPath, 0777) // ignore err?
```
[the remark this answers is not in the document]
TripleDogDare: :shrug: We can remove it and see how it goes.

cmd/warpforge/watch.go
TripleDogDare: ehhh... fine.

TripleDogDare: My primary reason for this would be that we might create other kinds of sockets instead of unix sockets.

cmd/warpforge/watch.go
TripleDogDare: Yes, the file locking was being used for the same purpose as this function. Would file locks actually make it _better_? Probably not in any significant way. I think the underlying coordination problem would still exist but just present in a slightly different way and possibly require user intervention to unlock in edge cases. So no; that's worse.

pkg/cmd/watch/watch.go
```diff
@@ -0,0 +1,394 @@
...
+				}
+			}
+			innerSpan.End()
+		}
+		outerSpan.End()
+		time.Sleep(time.Millisecond * 100)
+	}
+}
+
+// DEPRECATED: This should be refactored significantly in the near future to remove reliance on PWD in plot exec.
```
TripleDogDare, line 329: This is being worked on _currently_ but out-of-scope for this PR.

cmd/warpforge/watch.go
warpfork: I can see other kinds of sockets being in the future, for sure.  But I still think this one also defaults to on, definitely for now, and probably in the future too even if there are also others.

pkg/cmd/watch/watch.go
```diff
@@ -0,0 +1,394 @@
...
+		log.Info("", "socket handler: %s", err.Error())
+		return err
+	}
+	return nil
+}
+
+// serve will call Accept and block until the listener is closed.
+// context will only be checked between accepted connections
+// serve should not return under normal circumstances, however if an error occurs during accept then it will log that error and return it
+// will log and return immediately if listen was not called
```
warpfork, line 88: Slightly broken formatting and sentences here, maybe just not quite finished?

warpfork:
I think I'm basically good with this.

In package naming: well, it's a bikeshed, but fwiw, I think `pkg/cmd/watch/` is slightly unnecessary in depth, and `cmd` is outright surprising to see anywhere except in something that's gonna contain main packages.  `pkg/watch/` would be fine by me.  Or at least some other grouping word there other than `cmd` would be preferable.  (Other opinions welcome.)

Some of the concurrency is getting a little eye-watering.  I'm going to assume it's right, but it's hard to eyeball it.  (It's not even that bad!  But golang has so little for convention about where channels go and how to wire them that it doesn't take much before it starts producing brain-strain.)  My suspicion is that I'm not going to be entirely happy until we get either a little bit of concurrency helper library, or a _lot_ of concurrency organization convention... but I'm not sure I can state what I'm dreaming of briefly right now, and I might just have to go write about that for a little bit to clarify it even to myself before making any asks in this department.

TripleDogDare:
>In package naming: well, it's a bikeshed, but fwiw, I think pkg/cmd/watch/ is slightly unnecessary in depth, and cmd is outright surprising to see anywhere except in something that's gonna contain main packages. pkg/watch/ would be fine by me. Or at least some other grouping word there other than cmd would be preferable. (Other opinions welcome.)

I couldn't really come up with anything reasonable for grouping what amounts to all our CLI business logic. But I do think that a grouping seems reasonable and putting everything directly under `pkg` seems like it will produce a lot of clutter there. It's a bit of a bikeshed. :shrug:

warpfork (approved):
:+1:

PR #43, opened 2022-12-06
handle workspace error on ingest

TripleDogDare:
unhandled error causing a nil reference panic


warpfork:
I'm not entirely sure the error code meaningfulness is as high as the ideal might be.

(Questions that shine a light on it might be: Is the error code something another part of application logic could meaningfully switch on, and react to?)

But if it's your assertion that this is better-than-before, :+1: and incremental progress is progress.

TripleDogDare:
>Is the error code something another part of application logic could meaningfully switch on, and react to?

Maybe? We could have a function that just runs a bunch of Mkdirs in all the locations where it might expect something to exist. But also we might want to rethink the expectations of exec so that, for example, we've already set up a valid cache dir before we run and not rely on workspaces as a concept within exec functions.

warpfork:
Yep.  Just a note that we may have future refactors to do here, and a systematic review of our errors someday.  Not necessarily today.  But it looks like it might be building up.

PR #44, opened 2022-12-08
Implement `plan generate` for Warplark

ericevenchick:
This adds an initial implementation of the `plan` subcommand and allows for `plan generate`. This will require [warplark](https://github.com/warptools/warplark) which we currently do not bundle. With `warplark` on your path, you can now run the warplark examples with Warpforge:

```
warpforge -v plan generate examples/...
warpforge run examples/...
```


examples/500-cli/cli.md
```diff
@@ -742,3 +742,152 @@ linuxrc
...
+
+[testmark]:# (plan-generate/then-recursive/output)
+```
+```
+
+[testmark]:# (plan-generate/then-recursive/then-cat/script)
+```
+cat a/plot.wf
+cat b/plot.wf
+```
```
warpfork, line 874: I think if you want to report more than one file here, a neat trick may be to use `head */*.wf`, which will print out all the contents (since in this case they're short), and also happens to include some textual separators.

Although in a slightly bigger picture also I'm a little neutral how interesting it to have these two files shown which are identical :)

warpfork (approved):
Seems like a great start on some exciting new stuff!  :+1:

PR #45, opened 2022-12-08
Fix duplicate GHA actions

ericevenchick:
We previously did two builds on every push to a PR. This performs builds on pushes to PRs, and also on pushes to main/master branch.


warpfork (approved):
yaml is wonderful and this config file is very clear and obvious.  yes, it definitely makes sense to me why this diff would have the effect described.  very.

:disappointed: anyway let's merge it

PR #48, opened 2022-12-15
Sketch of possible API for local workspace situation reporting.

warpfork:
This commit is foremost an outline and proof-of-concept that we can split definitions of our schema into multiple files, and reuse the main warpforge API schema types in this second, superset schema.  And yep: that works.  (I did something a tad hacky to achieve it.  I'll try to push at least some of that towards upstream.  It's mild, though.)

The actual message types declared in this diff are a sketch only, and not firmly committed to.

### DIFF
--- a/pkg/workspaceapi/api.go
+++ b/pkg/workspaceapi/api.go
@@ -1,0 +1,53 @@
+package workspaceapi
+
+import (
+	"embed"
+	"fmt"
+
+	"github.com/warptools/warpforge/wfapi"
+
+	"github.com/ipld/go-ipld-prime/node/bindnode"
+
+	"github.com/ipld/go-ipld-prime/datamodel"
+
+	"github.com/ipld/go-ipld-prime/schema"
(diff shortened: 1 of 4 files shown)


pkg/workspaceapi/api.go
```diff
@@ -0,0 +1,53 @@
...
+	if err := schemadmt.Compile(ts, schemaDmt); err != nil {
+		panic(fmt.Sprintf("failed to compile api schema: %s", err))
+	}
+	return schemaDmt, ts
+}()
+
+// concat returns a new schemadmt that's got the types from both.
+//
+// This function could probably be hoisted upstream.
+func concat(a, b *schemadmt.Schema) *schemadmt.Schema {
```
TripleDogDare, line 44: This function should return an error and have the anonymous init function above do the panics.

TripleDogDare (approved):
We'll probably make changes to this but it has the base ipld code that I'll need to get started.
Minor change request that has no real effect on execution so :shrug:. LGTM.

TripleDogDare:
This will be included in #58

PR #50, opened 2022-12-25
Fixes partial bundle error

TripleDogDare:
Bundling catalogs with multiple items from a release would error on the second item because warpforge would be looking in the local workspace version of that release which would only contain the first item. This change allows continuing up the workspace stack to find the item.


TripleDogDare:
I'm not super sure how to write tests for this. But we may want to consider copying whole releases instead for bundle behavior.

warpfork (approved):
- 👍 agree to that error code rename towards more subsystem nouns first.
- agree that more tests desired around this in the future...
- this seems incrementally better, LGTM

PR #51, opened 2022-12-27
replace wfapi.Error with serum errors

TripleDogDare:
This is a pretty significant consistency change so that we don't get stuck supporting a mix of wfapi.Error and serum errors. I've found that they interact in strange ways (particularly around handling "cause" and how they serialize). This should make other work a bit easier.

Updates serum to v0.8.0 for quoted templates (i.e. `{{thing | q}}`) which we are already using but wouldn't actually serialize properly if we ever hit them using v0.7.0... oops.


pkg/workspace/workspace.go
```diff
@@ -11,6 +11,11 @@ import (
 	"github.com/warptools/warpforge/wfapi"
 )
 
+type codedError interface {
```
warpfork, line 14: Shouldn't be necessary.  There's package scope `serum.Code` func that'll do the relevant job.

TripleDogDare: Ah nice.

wfapi/error.go
```diff
@@ -137,28 +71,27 @@ func ErrorSearchingFilesystem(searchingFor string, cause error) Error {
 // Errors:
 //
 //  - warpforge-error-invalid --
-func ErrorInvalid(message string, deets ...[2]string) Error {
-	return &ErrorVal{
-		CodeString: "warpforge-error-invalid",
-		Message:    message,
-		Details:    deets,
+func ErrorInvalid(message string, deets ...[2]string) error {
```
warpfork, line 74: I think for some of these we maybe can just not bother having special constructor functions at all.  I think it's sorta fine if we see the serum package's constructors being used across the codebase.

The biggest reason we had a constructor for every single error here before was that we had really no other syntactically reasonable way to avoid it.  Now I think using the serum package widely is fine.  Additional constructor functions in this package are now only useful if they can take structured parameters (either well-typed, or a series of things that should always be specified together and we want to use the function to remind us).

This line I'm commenting on is a particularly outstanding not-worth-it-anymore, but there's probably others that don't make the cut anymore too.

TripleDogDare: Seems reasonable. I'm not super prepped to be making those decisions tonight though.

warpfork: Right, I...wish I could deploy a small time machine to save you from already having written this bit, heh.

TripleDogDare: Honestly nothing stands out to me as egregiously bad.

warpfork:
Replaying from a chat we had in other platforms, here for posterity:

> I propose we do all the consts as `Ecode*`

> Bikeshed color thing

> Some deep neckbeardy part of me really wants the autocomplete sequence to be starting with 'E'

> Golang people tend to expect `Err*` to be full on error values -- so we should not do that name pattern for these consts, which aren't full error values

And that's about how we got to this convention.  It seems to satisfice.

warpfork (approved):
LGTM

PR #53, opened 2023-01-03
Internal setup

TripleDogDare:
This is a draft of a pretty significant refactor where I try to wrangle and centralize our global state.


cmd/warpforge/internal/util/util.go
```diff
@@ -142,9 +141,8 @@ func ModuleFromFile(filename string) (wfapi.Module, error) {
 //    - warpforge-error-plot-invalid -- when the plot data is invalid
 //    - warpforge-error-plot-step-failed --
 //    - warpforge-error-serialization -- when the module or plot cannot be parsed
-//    - warpforge-error-unknown -- when changing directories fails
```
warpfork, line 145: this delete is already music to my ears :)

TripleDogDare:
@warpfork FWIW, there are two main sections in here. The first is the `pkg/config` package itself and the second is what I did to `formulaexec`. Pretty much everything that follows after that is a natural conclusion and stripping out stuff as I find it.

pkg/config/exec_config.go
```diff
@@ -0,0 +1,125 @@
...
+	return e.workspaceSet
+}
+
+// NewConfig returns a warpforge configuration
+//
+// Errors:
+//
+//   - warpforge-error-initialization -- unable to load value
+//   - warpforge-error-searching-filesystem -- unable to load workspace stack
+func NewConfig() (ExecConfig, error) {
```
warpfork, line 56: Would like to have this function take more parameters.  (I know this PR is an incremental improvement on gathering a bunch of this stuff from happening in many places to just one, but I'd also kinda like to take this one step further while we're here, and have less ambient magic happening too.)

Seems like the fsys and cwd as a parameter, are the two biggest ones.  A `getEnv func(string) string` or `envOverride map[string]string` might be a good idea too.

Or: if the parameter list seems like it's then getting too long and unwieldy: okay, fine; then let's put them in another struct, and have some sort of Init function on it that does the ambient-read stuff (e.g. all the Getwd and GetEnv stuff).

Any design that has an entrypoint that's got no ambient reads is a thumbs up for me.  It can have a zero-arg super-magic function that does all the ambient reads too; I just don't want to be stuck with that as the only entrypoint.

warpfork (commented):
Lotta questions about this guarded value thing.  Can we do without it?

pkg/formulaexec/formula_exec.go
```diff
@@ -195,25 +203,72 @@ func getBaseConfig(ctx context.Context, wsPath, runPath, binPath string) (runCon
...
+		interactive: false,
+	}
+	spec, err := copySpec(baseSpec)
 	if err != nil {
+		return rc, err
+	}
+	rc.spec = spec
+
+	// set up root -- this is not actually used since it is replaced with an overlayfs
+	rootPath := filepath.Join(runPath, "rooti") // FIXME: "root" is created as a regular file in the .warpforge dir? I think I screwed something up.
```
warpfork, line 269: rooti?  And is this fixme comment still living?

pkg/formulaexec/formula_exec.go
```diff
@@ -48,6 +48,20 @@ type RioOutput struct {
 	Result RioResult `json:"result"`
 }
 
+// DefaultRunPathPrefix will be the prefix used to create a temporary execution directory
+const DefaultRunPathPrefix = "warpforge-run-"
+
+type ExecConfig interface {
```
warpfork, line 54: Why'd this end up needing to declare a new interface?  Can't we just concretely import the new config package?

pkg/config/exec_config.go
TripleDogDare: Adding an `FS() fs.FS` might be a good idea. But consumers can choose what parts of the interface they want or implement new ones, so overrides feel somewhat redundant. I'd want an example of say `formulaexec.Exec` _taking_ an interface with an `FS()` method and using it before implementing it.

pkg/formulaexec/formula_exec.go
TripleDogDare: hmmm well we'll find out if it explodes when I remove the `i`.

pkg/config/setup.go
```diff
@@ -0,0 +1,199 @@
...
+
+type gvi interface {
+	// Errors:
+	//
+	//   - warpforge-error-initialization -- loading the value failed
+	load() error
+	reset()
+}
+
+type guardedValue[T any] struct {
```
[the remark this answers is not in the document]
TripleDogDare: Yeah this entire thing was a few hours of ADHD hyperfocus and possibly quite questionable. The `wip` commit shows what everything was like the day prior which is radically more in-line with what you probably would have built from a philosophical standpoint.

pkg/config/setup.go
```diff
@@ -0,0 +1,199 @@
...
+
+// load will execute the value fetcher but does not return the value
+//
+// Errors:
+//
+//   - warpforge-error-initialization -- loading the value failed
+func (g *guardedValue[T]) load() error {
+	if g.once == nil {
+		g.reset()
+	}
```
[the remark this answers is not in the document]
TripleDogDare: Quite definitely yes on both counts.

pkg/config/setup.go
```diff
@@ -0,0 +1,199 @@
...
+	g.once.Do(func() {
+		v, err := g.fetcher()
+		g.value = v
+		g.err = err
+	})
+	if g.err != nil {
+		return serum.Error(wfapi.ECodeInitialization,
+			serum.WithMessageTemplate("error fetching guarded value {{name | q}}"),
```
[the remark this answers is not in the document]
TripleDogDare: Going to remove all this.

pkg/plumbing/config/exec_config.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+func PlotExecConfig(cfg *ExecConfig) *plotexec.ExecConfig {
+	return &plotexec.ExecConfig{
+		BinPath:          cfg.BinPath(),
+		KeepRunDir:       cfg.KeepRunDir(),
+		RunPathBase:      cfg.RunPathBase(),
+		WhPathOverride:   cfg.WarehousePathOverride(),
+		WorkingDirectory: cfg.WorkingDirectory(),
+		WorkspaceSet:     cfg.WorkspaceSet,
```
warpfork, line 98: So I'm must wonder out loud here: how many fields being flipped from struct to struct is it before it begins to be a smell?  It seems like here (and the block below) there's more fields being copied than fields that are novel in the new struct.

I'd seriously consider just joining more things together in one big struct.  A heuristic I'm using for how much I like this code is "if we find there's another piece of config to track across the system, how many lines am I going to have to add?" and "if I rename something, how many sites have the same name and deserve the same change, while being technically different sites?", and these struct copies seem likely to make both those numbers go up.  Those numbers going up mean ease of change is decreasing.

In the other direction, putting more things into one bulky struct is creating closer coupling, but... I don't think we really have a compelling reason to treat that as a negative in this situation.  We really do use approximately all of these values approximately everywhere in Warpforge, afaict.  Or to put it another way: I currently have no goalposts in mind where things would be sufficient not-close-coupled to be a worthy target -- so nudging in that direction, without a decisive goalpost, doesn't seem worth it to me.

pkg/plumbing/config/exec_config.go
```diff
@@ -0,0 +1,113 @@
...
+	_, result := c.State.Env[key]
+	return result
+}
+
+func (c *ExecConfig) envOverride(key string) (string, bool) {
+	a, b := c.State.Env[key]
+	return a, b
+}
+
+func PlotExecConfig(cfg *ExecConfig) *plotexec.ExecConfig {
```
warpfork, line 91: The config package importing one of the thing-doing packages is the opposite direction I'd expect the package relationships to go.  A config package is usually going to land at the deepest parts of the package import graph.  So having it import much else, in my experience, usually provokes a cycle.

Apparently it hasn't produced a cycle yet, if this compiles successfully -- but I'd bet several shiny nickles going this direction will produce a cycle sooner or later as the code grows.

TripleDogDare: Importing global state into thing-doing packages is precisely what we are trying to avoid. If this causes a cycle then you have in fact imported global state and should redefine your package API.

pkg/plumbing/config/exec_config.go
TripleDogDare: I tend to care less about how many lines might change and more about whether or not I can track inputs from clearly defined API points. Renaming some fields is one of the least annoying things in the set of annoying things related to writing software. This particular bit of code is still a bit questionable just because it's superfluous. Where we could handle it via the utility functions.

pkg/plumbing/config/exec_config.go
TripleDogDare: It might be useful to see how I think the import graph should look.
imports

warpfork: If you have the iterated version of this diagram and wanna post that here too, I think it did get much improved over time :)

pkg/config/exec_config.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	"os"
+	"path/filepath"
+
+	"github.com/warptools/warpforge/pkg/formulaexec"
+	"github.com/warptools/warpforge/pkg/plotexec"
+	"github.com/warptools/warpforge/pkg/workspace"
+)
+
+func BinPath(state State) string {
```
warpfork, line 12: Not really clear to me why this takes State as a parameter rather than being a method

pkg/config/exec_config.go
```diff
@@ -0,0 +1,67 @@
...
+		return nil
+	}
+	return &value
+}
+
+// Retrieves workspace stack at state.WorkingDirectory
+// Errors:
+//
+//  - warpforge-error-searching-filesystem -- unexpected error traversing filesystem
+func DefaultWorkspaceStack(state State) (workspace.WorkspaceSet, error) {
```
warpfork, line 46: Think I'd probably make this a sibling of the `workspace.FindWorkspaceStack` in that package, while still accepting a config object as a parameter.

pkg/config/exec_config.go
```diff
@@ -0,0 +1,67 @@
...
+
+func FormulaExecConfig(state State) formulaexec.ExecConfig {
+	return formulaexec.ExecConfig{
+		BinPath:          BinPath(state),
+		KeepRunDir:       KeepRunDir(state),
+		RunPathBase:      RunPathBase(state),
+		WhPathOverride:   WarehousePathOverride(state),
+		WorkingDirectory: state.WorkingDirectory,
+	}
+}
```
warpfork, line 67: This seems massively redundant.

pkg/config/setup.go
```diff
@@ -0,0 +1,169 @@
...
+	Things like env vars and pwd can change during runtime.
+	This is a highly probable source of undefined or confusing behavior.
+	It may even force strange behavior such as moving directories temporarily
+	before running an operation, which would preclude concurrent operations.
+	This may not be the _most correct_ or _best_ way to handle these issues.
+	But it is certainly a step forward.
+	This doesn't prevent changes that might be passed down to sub-procsses like rio, runc
+*/
+
+type State struct {
```
warpfork, line 25: Propose calling this struct `Config` or something similar.

`config.Config` is a stutter, but that's merely unfortunate.

"State" doesn't really seem accurately descriptive.  None of these values are really meant to be seen as mutable.

warpfork (commented):
I still have a lot of comments about the config part.

All the other refactoring parts are basically thumbs up.

The config could be a lot more KISS.

pkg/plumbing/config/exec_config.go
TripleDogDare: architecture

Here's the updated version based.

pkg/config/module.go
```diff
@@ -0,0 +1,65 @@
...
+//
+// If no module file is found, it will return nil for the error value.
+// If errors are returned, they're due to filesystem IO.
+//
+// An fsys handle is required, but is typically `os.DirFS("/")` outside of tests.
+//
+// Errors:
+//
+//    - warpforge-error-searching-filesystem -- when an unexpected error occurs traversing the search path
+func FindModule(fsys fs.FS, basisPath, searchPath string) (path string, remainingSearchPath string, err error) {
```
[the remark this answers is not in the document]
TripleDogDare: Eh this file is no longer relevant.

pkg/config/exec_config.go
TripleDogDare: Dunno. Personal whims at the time.

pkg/config/setup.go
```diff
@@ -0,0 +1,169 @@
...
+	Env              map[string]string
```
warpfork, line 26: I think it would be more consistent to flatten the contents of this map into more fields rather than a map.

It would only be four more fields.

(Or if we end up deciding gathering these values into a struct doesn't provide significant value at all, and we just do package scope funcs for each thing: then disregard this comment; that approach probably naturally ends up with no env map either.)

pkg/config/exec_config.go
TripleDogDare: I'm going to remove it and leave it up to packages to create a "pwd+wss" utility function where needed.

pkg/config/setup.go
TripleDogDare: eh fine removed.

pkg/config/setup.go
TripleDogDare: removed struct.

pkg/config/exec_config.go
TripleDogDare: It is now... less redundant? Kinda?

cmd/warpforge/internal/util/util.go
```diff
@@ -94,51 +107,55 @@ func PlotFromFile(filename string) (wfapi.Plot, error) {
...
 		return result, err
 	}
-
-	plot, werr := dab.PlotFromFile(fsys, filepath.Join(filepath.Dir(fileName), dab.MagicFilename_Plot))
-	if werr != nil {
-		return result, werr
+	modulePath := filepath.Dir(fileName)
+	modulePath, xerr := filepath.Abs(modulePath)
+	if xerr != nil {
+		return result, serum.Error(wfapi.ECodePathAbs, serum.WithCause(xerr),
```
warpfork, line 128: Gonna pick on this error code because I see it only comes up once, here: can we reuse some other existing error code for it?  Or perhaps get rid of this error branch entirely?

`filepath.Abs` is doing an `os.Getwd` internally.  That's the only possible source of its errors.  If we already have cwd somewhere, we can just do the clean+ifAbs+join ourselves and it would have no error branch.

warpfork (approved):
Found one more small improvable bit about error handling, but otherwise, this LGTM now.  Many small improvements here to land.  A bunch of config is now more all in one place, and that's progress.  A lot may also yet be getting future iterations, but landing this set and doing more PRs now seems good.

cmd/warpforge/internal/util/util.go
TripleDogDare: That's good to know. Probably a potential source of bugs in all sorts of places.

TripleDogDare: We've been using `ECodeIO` in a bunch of places for this. I'm not a fan of that either but we can maybe phase it out.

warpfork: I think `ECodeInitialization` in some places now too.  I'd probably lean towards that one (and use ECodeIO more for things that error while in motion).  (Weakly held opinion.)

warpfork (approved):

pkg/plotexec/plot_exec.go
```diff
@@ -193,6 +193,22 @@ func plotInputToFormulaInputSimple(ctx context.Context,
 		}
 
 		if wareId == nil {
+			logger.Debug(LOG_TAG, "failed to resolve catalog reference to ware ID: %s", basis.CatalogRef.String())
+			for _, ws := range wss {
```
TripleDogDare, line 197: note: removing this chunk of debug logging

pkg/workspace/catalog.go
```diff
@@ -186,10 +189,12 @@ func (cat *Catalog) GetRelease(ref wfapi.CatalogRef) (*wfapi.CatalogRelease, err
 
 	// release was found in catalog, attempt to open release file
 	releasePath := cat.releaseFilePath(ref)
+	fmt.Println(releasePath)
```
TripleDogDare, line 192: note: removing this

PR #55, opened 2023-01-04
DRAFT: add warplark via go.mod tooling

TripleDogDare:
You may have mixed feelings about this. But it is an option that isn't committing the binary directly to plugins.


Makefile
```diff
@@ -20,6 +21,26 @@ endif
...
 
+warplark: check-warplark
+ifndef WARPLARK
+	@echo "Building and installing warplark..."
+	go install github.com/warptools/warplark
+endif
+
+update-warplark:
+	go install github.com/warptools/warplark@latest
+	sha256sum $$(which warplark) | cut -f1 -d' ' > tools/warplark.sha256
```
TripleDogDare, line 32: this is dumb. the hash will change based on which go compiler you are using.
We would have to pin the compiler version as well.

PR #56, opened 2023-01-05
Add validation rules to module names

TripleDogDare:
Also refactors some utility functions by placing them into the dab package. Overall this helps ensure that modules are being validated by centralizing the module file handling.

### DIFF
--- a/cmd/warpforge/check.go
+++ b/cmd/warpforge/check.go
@@ -3,13 +3,17 @@ package main
 import (
 	"context"
 	"fmt"
-	"io/ioutil"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"strings"
 
 	"github.com/ipld/go-ipld-prime"
 	"github.com/ipld/go-ipld-prime/codec/json"
 	"github.com/urfave/cli/v2"
 
 	"github.com/warptools/warpforge/cmd/warpforge/internal/util"
+	"github.com/warptools/warpforge/pkg/dab"
 	"github.com/warptools/warpforge/pkg/plotexec"
 	"github.com/warptools/warpforge/wfapi"
 )
@@ -24,8 +28,8 @@ var checkCmdDef = cli.Command{
 	),
 }
 
-func checkModule(fileName string) (*ipld.Node, error) {
-	f, err := ioutil.ReadFile(fileName)
+func checkModule(fsys fs.FS, fileName string) (*ipld.Node, error) {
+	f, err := fs.ReadFile(fsys, fileName)
 	if err != nil {
 		return nil, wfapi.ErrorIo("cannot read module file", fileName, err)
 	}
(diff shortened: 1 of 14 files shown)


pkg/dab/module.go
```diff
@@ -15,6 +19,164 @@ const (
...
+	validation_dns1123Subdomain_msg         string = "must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character"
+	validation_dns1123Subdomain_maxlen      int    = 253
+)
+
+var (
+	validation_moduleNamePathHunk_regexp = regexp.MustCompile("^" + validation_moduleNamePathHunk_regexpStr + "$")
+	validation_dns1123Subdomain_regexp   = regexp.MustCompile("^" + validation_dns1123Subdomain_regexpStr + "$")
+)
+
+// ValidateModuleName checks the module name for invalid strings.
```
Morganamilo, line 41: Maybe add some examples off valid module names in this comment.

cmd/warpforge/check.go
```diff
@@ -78,34 +89,43 @@ func cmdCheck(c *cli.Context) error {
 		return fmt.Errorf("no input files provided")
 	}
 	ctx := c.Context
-
+	pwd, err := os.Getwd()
+	if err != nil {
+		return err
```
Morganamilo, line 94: Should this return a wapi error? Something like `return serum.Errorf(wfapi.ECodeUnknown, "unable to get pwd: %w", err)`

Morganamilo (commented):
Looks good to me from an uneducated viewpoint. Only a few nits.

cmd/warpforge/check.go
TripleDogDare: Yeah. Serum doesn't detect these so they are pretty inconsistent across everything in `cmd/*`. This should improve as we move things under the plumbing package since we'll have more consistent command setup/execution and the business logic entry point will be an exported function that serum checks.

TripleDogDare: Adding an `Errors:` comment to our command entry points would coerce serum into checking these functions which would probably be a good idea for a separate PR.

pkg/dab/module.go
TripleDogDare: Hmm... perhaps. I'm not convinced it will add value but it's not a lot of effort and I need to change this doc string anyway.
I definitely need to remove the line about underscores being allowed since that's no longer true!

warpfork: If there's anything I value more and more over the years, it's examples.  :)

If in the docs doesn't feel robust enough, the next two steps up are: golang's `Example_` test functions are pretty neat for getting things into godoc; or, making a quick invocation of testmark and putting it in our markdown docs directories is also almost always great.  (I'd say in this case, the latter is more likely to be useful -- we'll probably expect end users to want to see examples of strings that are recommended for use here, and they won't all be godoc consumers.)

pkg/dab/module_test.go
```diff
@@ -0,0 +1,175 @@
...
+		{"", "example.com/b/c", qt.IsNil},
+		{"", "1.2", qt.IsNil},
+		{"", "1.2/3", qt.IsNil},
+		{"", "1.2/3/4", qt.IsNil},
+		{"", "a.b/foo", qt.IsNil},
+		{"", "a.b/foobar", qt.IsNil},
+		{"", "a.b/foo/bar", qt.IsNil},
+		// underscore considered for valid chars but removed
+		{"", "abc_def.ghi/jkl", qt.IsNotNil},
+		{"", "a.b/foo_bar", qt.IsNotNil},
```
warpfork, line 42: I wouldn't mind putting underscore into the middle grouping of `validation_moduleNamePathHunk_regexpStr`, which would make `"a.b/foo_bar"` valid.  I don't feel strongly about it, but I feel like restricting that char might produce more headdesk than not.

(I think we already have plenty of space reserved for all other purposes, e.g. the filesystem projection of catalogs, already from reserving leading underscore alone.)

warpfork (approved):
Looks pretty good to me!

One tiny range expansion I'd vote for.  If taken, the tests are already right there for it and everything, so that's nice and makes it easy :D

If we ended up with more example fixtures that make it all the way out to docs, that'd be nice, but that can also be future work.

TripleDogDare:
I'll probably merge this tomorrow (approximately +18h) unless someone adds any further comments.

PR #57, opened 2023-01-06
Use .warphome as the name for the home workspace

Morganamilo:
This stops is from being possible to acceidently get some one else's home workspace as part of your workspace stack.

### DIFF
--- a/pkg/workspace/workspace.go
+++ b/pkg/workspace/workspace.go
@@ -49,14 +49,20 @@ func OpenWorkspace(fsys fs.FS, rootPath string) (*Workspace, error) {
 // from an outside perspective.
 func openWorkspace(fsys fs.FS, rootPath string) *Workspace {
 	rootPath = filepath.Clean(rootPath)
-	isHomeWorkspace := rootPath == homedir
 	return &Workspace{
 		fsys:            fsys,
 		rootPath:        rootPath,
-		isHomeWorkspace: isHomeWorkspace,
-		isRootWorkspace: checkIsRootWorkspace(fsys, rootPath) || isHomeWorkspace,
+		isRootWorkspace: checkIsRootWorkspace(fsys, rootPath),
 		// that's it; everything else is loaded later.
 	}
+}
+
+// openHomeWorkspace is the same as the public method but with no error checking at all;
+func openHomeWorkspace(fsys fs.FS) *Workspace {
+	workspace := openWorkspace(fsys, homedir)
+	workspace.isHomeWorkspace = true
+	workspace.isRootWorkspace = true
+	return workspace
 }
 
 // OpenHomeWorkspace calls OpenWorkspace on the user's homedir.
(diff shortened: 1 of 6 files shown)


pkg/workspace/workspace.go
```diff
@@ -114,8 +133,7 @@ func (ws *Workspace) IsRootWorkspace() bool {
 func (ws *Workspace) MemoBasePath() string {
```
TripleDogDare, line 133: Change the function comment from `i.e.` to `e.g.`

pkg/workspace/workspace.go
```diff
@@ -131,15 +149,20 @@ func (ws *Workspace) MemoPath(fid string) string {
 // Returns the base path which contains named catalogs (i.e., `.../.warpforge/catalogs`)
 func (ws *Workspace) CatalogBasePath() string {
```
TripleDogDare, line 150: Change the function comment from `i.e.` to `e.g.`

pkg/workspace/workspace.go
```diff
@@ -68,11 +74,25 @@ func openWorkspace(fsys fs.FS, rootPath string) *Workspace {
...
+		workspace.isRootWorkspace = true
+	}
+	return workspace, err
+}
+
+// DotPath returns the workspace's path including the .warp* segment.
+func (ws *Workspace) DotPath() string {
+	if ws.isHomeWorkspace {
+		return filepath.Join(ws.rootPath, magicHomeWorkspaceDirname)
+	} else {
```
TripleDogDare, line 89: `else` is redundant due to early return pattern.

warpfork (approved):
I think we all basically agreed on this already at the high level, so I'll give this a very quick LGTM.

I think I'm also going to go drop a tag now though, right before we merge this.  (We're not taking versioning super seriously at this stage of development, but this change still feels like a pretty noticeable bump.)  I'll tag a `v0.6.0` right before this, and I figure it might be reasonable to tag a `v0.7.0` fairly immediately after this.

pkg/workspace/workspace.go
```diff
@@ -68,11 +74,25 @@ func openWorkspace(fsys fs.FS, rootPath string) *Workspace {
...
+	workspace, err := OpenWorkspace(fsys, homedir)
+	if err == nil {
+		workspace.isHomeWorkspace = true
+		workspace.isRootWorkspace = true
+	}
+	return workspace, err
+}
+
+// DotPath returns the workspace's path including the .warp* segment.
+func (ws *Workspace) DotPath() string {
```
TripleDogDare, line 86: I'm not a huge fan of `DotPath` as a name because it describes the path format rather than what it does. Something like `InternalsPath` would be more descriptive of function.

TripleDogDare:
Got a few nits. Otherwise LGTM.

pkg/workspace/workspace.go
Morganamilo: Good point, I'll rename it to this.

Morganamilo:
> I think we all basically agreed on this already at the high level, so I'll give this a very quick LGTM.
> 
> I think I'm also going to go drop a tag now though, right before we merge this. (We're not taking versioning super seriously at this stage of development, but this change still feels like a pretty noticeable bump.) I'll tag a `v0.6.0` right before this, and I figure it might be reasonable to tag a `v0.7.0` fairly immediately after this.

Okay, I'll leave it to you to merge then.

PR #58, opened 2023-01-13
CI spark

TripleDogDare:
Adds some features to CI functionality.
 - A step closer to workspace based CI.
 - An experimental workspace API.
 - Actual RPC communication over socket using said API.
 - Features to put status codes into text-based systems (bash prompt, i3 blocks, etc.)
 - A very small push towards an actor model for the watch command.

Not in this PR:
 - Full workspace/Multi-module watch.
 - Setting a default module for a workspace.

Requesting status:
image
Changing to a passing branch and requesting status again.
image

Notes:
1) Why "spark"? It's what @warpfork and I called something with similar functionality. In fact, some of the output formatting code comes from there directly.
2) Where to start review? This is pretty hefty but probably the workspaceapi schema. Here's the general breakdown with line estimates at time of writing:
  - (115 lines, ~4%) cli changes under `cmd/...`
  - (86 lines, ~3%) wfapi changes under `wfapi/...` mostly error codes and examples for ware ID handling.
  - (983 lines, ~33%) server is under `pkg/plumbing/watch/...`. This builds on top of existing `watch` command.
  - (660 lines, ~22%) client is under `pkg/plumbing/spark/...`. This is entirely new stuff. The old "client" was just netcat.
  - (494 lines, ~17%) api is under `pkg/workspaceapi/...`. New! This is mostly schema and you should probably go over it before server/client stuff.
  - (991 lines, ~33%) Major changes to workspace detection. Mostly tests. I ended up adding a bunch of tests because it wasn't behaving the way I expected it to. This probably deserves some inspection independently of the server/client interactions. This is split off into it's own PR #70. So we should probably get that reviewed first.
  - (83 lines, ~3%) dab.FindModule: likely to be superseded by #69.


TripleDogDare:
Includes most of api spec from #48

pkg/workspaceapi/api.go
```diff
@@ -0,0 +1,178 @@
...
+type ModuleStatusUnion struct {
+	ModuleStatusUnion_NoInfo             *ModuleStatusUnion_NoInfo
+	ModuleStatusUnion_Queuing            *ModuleStatusUnion_Queuing
+	ModuleStatusUnion_InProgress         *ModuleStatusUnion_InProgress
+	ModuleStatusUnion_FailedProvisioning *ModuleStatusUnion_FailedProvisioning
+	ModuleStatusUnion_ExecutedSuccess    *ModuleStatusUnion_ExecutedSuccess
+	ModuleStatusUnion_ExecutedFailed     *ModuleStatusUnion_ExecutedFailed
+}
+
+func UnionField(i interface{}) string {
```
warpfork, line 145: You can also just get this by looking at it with IPLD stuff and asking the map key.

Unions when viewed at type level always act like a map with one key, and the key is the name of the member type.

TripleDogDare: Positive: less reflect (or just hiding it behind the IPLD lib). Negative: a bit weird. Maybe there's a better way to do it that I'm not seeing.
 
```golang
// Errors:
//
//   - warpforge-error-unknown -- Unable to extract union key
func (r *RpcRequest) Kind() (string, error) {
	sch := TypeSystem.TypeByName("RpcRequest")
	return unionField(r, sch)
}

// Errors:
//
//   - warpforge-error-unknown -- Unable to extract union key
func unionField(i interface{}, sch schema.Type) (string, error) {
	n := bindnode.Wrap(i, sch)
	iter := n.MapIterator()
	key, _, err := iter.Next()
	if err != nil {
		return "", serum.Error(wfapi.ECodeUnknown, serum.WithCause(err))
	}
	result, err := key.AsString()
	if err != nil {
		return "", serum.Error(wfapi.ECodeUnknown, serum.WithCause(err))
	}
	return result, nil
}
```

pkg/workspaceapi/api.go
```diff
@@ -0,0 +1,178 @@
...
+	"github.com/ipld/go-ipld-prime/node/bindnode"
+	"github.com/ipld/go-ipld-prime/schema"
+	"github.com/serum-errors/go-serum"
+)
+
+const (
+	ModuleInterestLevel_Query ModuleInterestLevel = "Query"
+	ModuleInterestLevel_Run   ModuleInterestLevel = "Run"
+)
+const (
```
TripleDogDare, line 18: It might be questionable to put these new error codes here instead of in wfapi. 
TODO: Needs description comments.

pkg/plumbing/watch/watch.go
```diff
@@ -218,65 +164,71 @@ func canonicalizePath(pwd, path string) string {
...
+		return err
 	}
-
-	modulePath := filepath.Join(c.Path, dab.MagicFilename_Module)
-	modulePathAbs := canonicalizePath(wd, modulePath)
-
-	// TODO: currently we read the module/plot from the provided path.
-	// instead, we should read it from the git cache dir
-	plot, err := dab.PlotFromFile(os.DirFS(c.Path), dab.MagicFilename_Plot)
+	_, wsPath := ws.Path()
```
ericevenchick, line 178: `FindWorkspace` returns nil if no workspace is found. This will cause a nil pointer deref.

pkg/plumbing/spark/spark.go
```diff
@@ -0,0 +1,230 @@
...
+	modulePath, _, err := dab.FindModule(c.Fsys, wsPath, searchPath)
+	if err != nil {
+		return empty, serum.Error(ECodeSparkNoModule, serum.WithCause(err))
+	}
+
+	query := workspaceapi.ModuleStatusQuery{
+		Path:          modulePath,
+		InterestLevel: workspaceapi.ModuleInterestLevel_Query,
+	}
+	logger.Info("", "ModulePath: %s", query.Path)
```
ericevenchick, line 83: This should likely be a `logger.Debug`, currently it appears by default when running spark.

ericevenchick: Also, missing leading `/`

pkg/plumbing/watch/server.go
```diff
@@ -0,0 +1,290 @@
...
+func NextRPC(ctx context.Context, d *json.Decoder) (*workspaceapi.Rpc, error) {
+	var err error
+	var raw json.RawMessage
+	var rpc workspaceapi.Rpc
+	log := logging.Ctx(ctx)
+	tag := handlerTag(ctx)
+	log.Debug(tag, "reading json")
+	err = d.Decode(&raw)
+	if err != nil {
+		return nil, serum.Error(workspaceapi.ECodeRpcSerialization, serum.WithCause(err),
```
ericevenchick, line 103: This error gets hit twice after every valid RPC. I believe we're hitting an EOF on `NextRPC` then another EOF when the connection is closed. Here is the full log for a single RPC request:

```
╬═  server[26]  getting next RPC
╬═  server[26]  handle request
╬═  server[26]  response
╬═  server[26]  getting next RPC
╬═  server[26]  unable to read RPC: warpforge-error-rpc-serialization: unable to read JSON from connection: caused by: bestguess-golang-errors-errorString: EOF
╬═  server[26]  connection closed
╬═  server[26]  handler returned with error: warpforge-error-rpc-serialization: unable to read JSON from connection: caused by: bestguess-golang-errors-errorString: EOF
```

ericevenchick (commented):
↯↯↯

Looks good with a few little nits. I think we do need to figure docs out for this feature though. It's somewhat unclear that you need an `ingest` to `watch` for instance.

I'm a little confused about the workspace part of this, right now it will return a nil pointer deref if there is no workspace (`.warpforge`). It also appears to ignore `.warphome`. It looks like there's dependence on both the working dir and the provided path.

pkg/plumbing/watch/watch.go
```diff
@@ -218,65 +164,71 @@ func canonicalizePath(pwd, path string) string {
...
-	for _, step := range plot.Steps.Values {
-		for _, input := range step.Protoformula.Inputs.Values {
-			allInputs = append(allInputs, input)
-		}
+	log.Debug("", "module path: %s", modulePath)
+	_, err = dab.ModuleFromFile(c.Fsys, modulePath)
+	if err != nil {
+		return err
 	}
+	moduleDir := filepath.Dir(modulePath)
```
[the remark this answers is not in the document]
TripleDogDare: I should be able to do a single `filepath.Join` after the call to FindModule.

pkg/plumbing/watch/watch.go
TripleDogDare: I had an old `.warpforge` dir in my home directory hiding this. :scream:

pkg/plumbing/watch/server.go
TripleDogDare: Yeah I've been ignoring that... I suppose I should handle it more gracefully.

TripleDogDare:
>It's somewhat unclear that you need an ingest to watch for instance.

I suspect that will change in the future but should be added.

> I'm a little confused about the workspace part of this, right now it will return a nil pointer deref if there is no workspace (.warpforge).

nil pointer ref will be fixed.

> It also appears to ignore .warphome.

Plot execution will get a full workspace stack. But it only watches modules that exist within a local workspace.

> It looks like there's dependence on both the working dir and the provided path.

This is generally true for most of our commands. I want users to be able to pass in whatever path they want (absolute or relative) and have things work as expected. I also want those things to be explicitly handed down the call stack so we don't have as many points where we need to handle errors from looking it up. And I definitely don't want to call `Chdir` in a command where we plan to be running concurrent operations.

pkg/dab/module.go
```diff
@@ -251,3 +252,45 @@ func PlotFromFile(fsys fs.FS, filename string) (wfapi.Plot, error) {
...
+//
+// If no module file is found, it will return nil for the error value.
+// If errors are returned, they're due to filesystem IO.
+//
+// An fsys handle is required, but is typically `os.DirFS("/")` outside of tests.
+//
+// Errors:
+//
+//    - warpforge-error-searching-filesystem -- when an unexpected error occurs traversing the search path
+func FindModule(fsys fs.FS, basisPath, searchPath string) (path string, remainingSearchPath string, err error) {
```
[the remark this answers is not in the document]
TripleDogDare: Yeah. It might be worth doing that when we get there.

pkg/plumbing/watch/watch.go
```diff
@@ -218,65 +164,71 @@ func canonicalizePath(pwd, path string) string {
...
 
-	for _, input := range allInputs {
-		if input.Basis().Ingest != nil && input.Basis().Ingest.GitIngest != nil {
-			ingest := input.Basis().Ingest.GitIngest
-			ingests[ingest.HostPath] = ingest.Ref
-		}
+	// TODO: currently we read the module/plot from the provided path.
+	// instead, we should read it from the git cache dir
+	plotPath := filepath.Join(moduleDir, dab.MagicFilename_Plot)
+	log.Debug("", "plot path: %s", plotPath)
```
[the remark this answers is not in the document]
TripleDogDare: Should be fixed.

pkg/plumbing/watch/watch.go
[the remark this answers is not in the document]
TripleDogDare: Should be fixed

pkg/plumbing/watch/watch.go
```diff
@@ -218,65 +164,71 @@ func canonicalizePath(pwd, path string) string {
...
-	for _, input := range plot.Inputs.Values {
-		allInputs = append(allInputs, input)
+	if modulePath == "" {
+		return serum.Error(wfapi.ECodeModuleInvalid, serum.WithMessageLiteral("no module found"))
 	}
-	for _, step := range plot.Steps.Values {
-		for _, input := range step.Protoformula.Inputs.Values {
-			allInputs = append(allInputs, input)
-		}
+	log.Debug("", "module path: %s", modulePath)
```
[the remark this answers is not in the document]
TripleDogDare: should be fixed

pkg/plumbing/watch/server.go
TripleDogDare: Server output should be debug only and it should ignore the EOF caused by the client dropping connections.

PR #60, opened 2023-01-20
Reduces duplication of errors

TripleDogDare:
Avoiding wrapping errors with serum.Errorf should reduce the duplication of error messages. Additionally marking a bunch of degenerate helper functions as deprecated for future removal. This commit strives to make few behavioral changes outside the primary purpose of reducing message duplication.

### DIFF
--- a/wfapi/error.go
+++ b/wfapi/error.go
@@ -2,7 +2,6 @@ package wfapi
 
 import (
 	"encoding/json"
-	"fmt"
 	"os"
 
 	"github.com/serum-errors/go-serum"
--- a/wfapi/error.go
+++ b/wfapi/error.go
@@ -7,44 +7,36 @@ import (
 	"github.com/serum-errors/go-serum"
 )
 
-// TODO: Add comments for reasons to use or not use particular codes
+// Error codes are loosely of the form "<application>-error-[subsystem]-<kind>".
 const (
-	// ECodeArgument may be used when invalid arguments are provided to the warpforge command line
-	ECodeArgument = "warpforge-error-invalid-argument"
-	// ECodeAlreadyExists may be used when _something_ already exists.
-	// Prefer to use a more specific error code or specify _what_ is missing.
-	ECodeAlreadyExists          = "warpforge-error-already-exists"
-	ECodeCatalogInvalid         = "warpforge-error-catalog-invalid"
-	ECodeCatalogMissingEntry    = "warpforge-error-catalog-missing-entry"
-	ECodeCatalogName            = "warpforge-error-catalog-name"
-	ECodeCatalogParse           = "warpforge-error-catalog-parse"
-	ECodeDataTooNew             = "warpforge-error-datatoonew"
-	ECodeExecutorFailed         = "warpforge-error-executor-failed"
-	ECodeFormulaExecutionFailed = "warpforge-error-formula-execution-failed"
-	ECodeFormulaInvalid         = "warpforge-error-formula-invalid"
-	ECodeGeneratorFailed        = "warpforge-error-generator-failed"
-	ECodeGit                    = "warpforge-error-git"
-	// ECodeInternal is used for errors that are internal and cannot be handled by users.
-	// Try to pick something more specific.
-	ECodeInternal = "warpforge-error-internal"
-	// ECodeInvalid is used when something is invalid.
-	// Prefer to choose a more specific error code.
-	ECodeInvalid             = "warpforge-error-invalid"
-	ECodeIo                  = "warpforge-error-io"
-	ECodeMissing             = "warpforge-error-missing"
-	ECodeModuleInvalid       = "warpforge-error-module-invalid"
-	ECodePlotExecution       = "warpforge-error-plot-execution-failed"
-	ECodePlotInvalid         = "warpforge-error-plot-invalid"
-	ECodePlotStepFailed      = "warpforge-error-plot-step-failed"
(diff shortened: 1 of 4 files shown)


wfapi/error.go
```diff
@@ -199,18 +192,20 @@ func ErrorFormulaInvalid(reason string) error {
...
 // formula execution to fail.
 //
 // Errors:
 //
 //    - warpforge-error-formula-execution-failed --
 func ErrorFormulaExecutionFailed(cause error) error {
-	return serum.Errorf(ECodeFormulaExecutionFailed,
-		"formula execution failed: %w", cause,
+	return serum.Error(ECodeFormulaExecutionFailed, serum.WithCause(cause),
+		serum.WithMessageLiteral("formula execution failed"),
```
warpfork, line 203: Agree with this one being deprecated.  I think this one is an example of where having a "cause" parameter at all was probably a misstep.

Formula execution does not have _that_ many moving parts.  Maybe this error code exists, and maybe it has a constructor function, but the constructor(s?) should probably be functions accepting some specific details as parameters... not an opaque, arbitrary cause.

wfapi/error.go
```diff
@@ -112,39 +110,38 @@ func ErrorWorkspace(wsPath string, cause error) error {
...
+	return serum.Error(ECodeExecutorFailed, serum.WithCause(cause),
+		serum.WithMessageTemplate("the {{engineName|q}} engine reported error"),
+		serum.WithDetail("engineName", executorEngineName),
 		// ideally we'd have more details here, but honestly, our executors don't give us much clarity most of the time, so... we'll see.
-	})
-	return result
+	)
 }
 
+// DEPRECATED: This is just adds a degenerate repetition of the error code
```
warpfork, line 119: Not sure if agree on this; it also nudges to attach a path.

(I'm not sure if all IO errors have this already and it's redundant; I think some don't?  I guess the ideal would be to only bother to attach the path on those types where it's not redundant, but this function just didn't grow to be that clever yet.)

wfapi/error.go
```diff
@@ -112,39 +110,38 @@ func ErrorWorkspace(wsPath string, cause error) error {
...
-	addDetails(result, [][2]string{{"context", context}, {"path", path}})
-	return result
+	return serum.Error(ECodeIo, serum.WithCause(cause),
+		serum.WithMessageTemplate("io error: {{context}}"),
+		serum.WithDetail("context", context),
+		serum.WithDetail("path", path),
+	)
 }
 
+// DEPRECATED: This is just adds a degenerate repetition of the error code
```
warpfork, line 133: I think this is not accurate; I'm pretty sure this is most usually being used to wrap errors from the ipld libraries (which aren't serum'ized).

(It's maybe not the most communicative either -- ideally it would say something about if it was during serialization, or deserialization, or struct mapping, or schema validation, or so on.  But I think this currently balks at that because it would be difficult to disambiguate some of these without either upstream patches, or building an awful lot of facades.)

wfapi/error.go
```diff
@@ -244,30 +240,29 @@ func ErrorMissingCatalogEntry(ref CatalogRef, replayAvailable bool) error {
...
 	}
 	return serum.Error(ECodeCatalogMissingEntry,
-		serum.WithMessageLiteral(msg),
+		serum.WithMessageTemplate(msg),
 		serum.WithDetail("catalogRef", ref.String()),
 		serum.WithDetail("replayAvailable", available),
 	)
 }
 
+// DEPRECATED: adds no value
```
warpfork, line 255: I think maybe this could be more helpful if it says what one should do instead.

`ECodeGit` is considered useful maybe?  And we want to say "just use serum.Error; there is no value-add from an additional constructor func"?  Is that what is meant?

warpfork: (Same goes for other occurrences of this comment further down.)

warpfork (commented):
Some comments on what's deprecated and how we communicate that.  I think this is mostly positive changes though.

warpfork:
Tested locally on some gnarly errors and this did make the messages much less redundant, so I think the desired effect is achieved.

wfapi/error.go
TripleDogDare: If we template the path into the message then I'd be more inclined to agree. Untemplated details only show up during serialization and not in the message user's see (unless you are in JSON mode). And some io errors not having paths makes it seem like we should have something more specific for io errors with paths. As it stands this just seems out-of-place and we should just call serum directly. We're mostly being saved by the fact that most io errors being wrapped happen to emit file paths.

wfapi/error.go
TripleDogDare: My point here is that It would be _just as valuable_ to wrap errors with `serum.Error(ECodeSerialization, serum.WithCause(cause), serum.WithMessageLiteral(context))`. And I don't feel that this meets the threshold for creating a helper function.

wfapi/error.go
TripleDogDare: >Is that what is meant?

Yes.

wfapi/error.go
TripleDogDare: The deprecation comment is pretty low-signal. These are more descriptive now.

wfapi/error.go
TripleDogDare: Deprecation comments are now more descriptive.

warpfork (approved):
This seems like quite a big improvement now.

(I don't love all of the error codes that we have right now!  Some of them I would probably like to split.  Some I would probably combine.  But documenting what we have is much better than not.)

PR #61, opened 2023-01-20
Mirroring (the pushing part)

ericevenchick:
This PR adds the `wf catalog mirror` command which puts wares into a remote warehouse. Currently, only S3-compatible hosting is supported, but other backends can be added.

This diff also makes a change to how catalog mirror files work. These used to be unions which could only contain a `byWare` or `byMirror` entry. Now, both are allowed. A specific WareID in `byWare` will resolve, but if none is found it will fallback to the `byModule` entry.

Workspaces can configure mirroring with a `mirroring.json` file. This sets the parameters for a specific `WarehouseAddr`. Currently this file needs to be manually created.

A branch of warpsys (https://github.com/warptools/warpsys/tree/use-byModule) exists that uses a `mirroring.json` and `byModule` entries. It is useful for testing this PR.


examples/500-cli/cli.md
```diff
@@ -616,7 +623,7 @@ warpforge --json run
 ```
 { "log": { "Msg": "inputs:" } } 
 { "log": { "Msg": "type = catalog ref = catalog:warpsys.org/busybox:v1.35.0:amd64-static" } } 
-{ "log": { "Msg": "wareId = tar:4z9DCTxoKkStqXQRwtf9nimpfQQ36dbndDsAPCQgECfbXt3edanUrsVKCjE9TkX2v9 wareAddr = none" } } 
+{ "log": { "Msg": "wareId = tar:4z9DCTxoKkStqXQRwtf9nimpfQQ36dbndDsAPCQgECfbXt3edanUrsVKCjE9TkX2v9 wareAddr = https://warpsys.s3.amazonaws.com/warehouse/4z9/DCT/4z9DCTxoKkStqXQRwtf9nimpfQQ36dbndDsAPCQgECfbXt3edanUrsVKCjE9TkX2v9" } }
```
warpfork, line 626: Does this imply that this test is regressing to fail in the absence of network?

pkg/mirroring/s3.go
```diff
@@ -0,0 +1,107 @@
...
+)
+
+type S3Pusher struct {
+	client       *s3.Client
+	cfg          wfapi.S3PushConfig
+	existingKeys map[string]bool
+}
+
+func wareIdToKey(wareId wfapi.WareID) string {
+	return filepath.Join(wareId.Hash[0:3], wareId.Hash[3:6], wareId.Hash)
```
TripleDogDare, line 24: As a TODO, we're doing this in a lot of places and maybe we should toss a default ware ID projection function somewhere.

pkg/mirroring/s3.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+func newS3Pusher(cfg wfapi.S3PushConfig) (S3Pusher, error) {
+	config, err := config.LoadDefaultConfig(context.TODO(),
```
TripleDogDare, line 28: I find that it is usually easier to add `context.Context` to the function APIs sooner rather than later and default to passing it along unless I'm fairly confident that a function is simple enough to not need network, logging, or async. I generally only use `context.TODO` to migrate an old codebase that isn't context aware (which is getting rarer these days).

pkg/mirroring/s3.go
```diff
@@ -0,0 +1,107 @@
...
+					URL:               cfg.Endpoint,
+					HostnameImmutable: true,
+					SigningRegion:     cfg.Region,
+				}, nil
+			})),
+		config.WithRegion(cfg.Region),
+	)
+
+	if err != nil {
+		panic(err)
```
TripleDogDare, line 42: Why a panic?

pkg/mirroring/mock.go
```diff
@@ -0,0 +1,25 @@
+package mirroring
+
+import "github.com/warptools/warpforge/wfapi"
+
+// A fake pusher that is intended for tests only. This will do nothing when "pushing" wares other than
+// keep track of the wares that have been pushed.
+
```
TripleDogDare, line 7: I think you need to remove the space for this to show up in go doc.

pkg/mirroring/push.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+func pusherFromConfig(cfg wfapi.WarehouseMirroringConfig) (pusher, error) {
+	if cfg.PushConfig.S3 != nil {
+		pusher, err := newS3Pusher(*cfg.PushConfig.S3)
+		return &pusher, err
+	} else if cfg.PushConfig.Mock != nil {
+		pusher, err := newMockPusher(*cfg.PushConfig.Mock)
+		return &pusher, err
+	} else {
```
TripleDogDare, line 30: Unnecessary else statement.

pkg/mirroring/push.go
```diff
@@ -0,0 +1,109 @@
...
+// for mirroring. The given catalog will be scanned for wares that can be pushed based on
+// the configuration.
+//
+// Errors:
+//
+// 	- warpforge-error-io -- for IO errors that occur during push operations
+//  - warpforge-error-catalog-invalid -- when the provided catalog contains invalid data
+//  - warpforge-error-catalog-missing-entry -- should never occur, as we iterate over the contents of the catalog
+//  - warpforge-error-catalog-parse -- when the provided catalog cannot be parsed
+func PushToWarehouseAddr(log *logging.Logger, ws workspace.Workspace, cat workspace.Catalog, pushAddr wfapi.WarehouseAddr, cfg wfapi.WarehouseMirroringConfig) error {
```
TripleDogDare, line 48: The logger should be configured in CLI layer and passed down in `context.Context`.

pkg/mirroring/s3.go
```diff
@@ -0,0 +1,107 @@
...
+
+}
+
+func (p *S3Pusher) hasWare(wareId wfapi.WareID) (bool, error) {
+	key := wareIdToKey(wareId)
+	if _, exists := p.existingKeys[key]; exists {
+		// key already exsits in bucket
+		return true, nil
+	} else {
+		return false, nil
```
TripleDogDare, line 83: Unnecessary else statement.

pkg/mirroring/push.go
TripleDogDare, line 30: Unnecessary else statement.

pkg/mirroring/push.go
```diff
@@ -0,0 +1,109 @@
...
+				}
+
+				// get the path to the data we want to push
+				warePath, _ := ws.WarePath(*wareId)
+
+				// it is possible we don't have this ware, in which case we just want to skip over it
+				if _, err := os.Stat(warePath); os.IsNotExist(err) {
+					log.Debug("mirror", "no local copy of wareId %q (expected at %q), skipping", wareId.String(), warePath)
+					continue
+				} else if err != nil {
```
TripleDogDare, line 86: I'm not a huge fan of this style of error handling with special cases. The scoping of `err` in `else if err != nil` is potentially ambiguous to humans who aren't particularly familiar with go. I understand that changing it means working around the limitations of the serum analyzer so it's kind of a toss-up on whether you want to do that. The easiest thing to do is just wrap it in brackets and hoist the `err := os.Stat` out of the if statement but inside the scoping brackets. Pretty ugly and I'm not sure I like that solution over possible solutions, but go has never been pretty.

go.mod
```diff
@@ -4,6 +4,10 @@ go 1.18
 
 require (
 	github.com/alecthomas/chroma v0.10.0
+	github.com/aws/aws-sdk-go-v2 v1.17.3
+	github.com/aws/aws-sdk-go-v2/config v1.18.7
+	github.com/aws/aws-sdk-go-v2/feature/s3/manager v1.11.46
+	github.com/aws/aws-sdk-go-v2/service/s3 v1.29.6
```
warpfork, line 10: Do you suppose you could collect little before-and-after numbers for what this does to the compiled binary sizes?

Features probably win over worrying about binary size -- but I'd like to be informed about the tradeoffs we're making.

(And yes, maybe feed this statistic forward into decisions about if we want to isolate these features into other binaries.  Whether today or in the future.)

wfapi/wfapi.ipldsch
```diff
@@ -430,4 +430,27 @@ type CatalogMirrorsByPacktype {Packtype:WarehouseAddrList}
...
+} representation keyed
+
+type S3PushConfig struct {
+	endpoint String
+	region String
+	bucket String
+	path optional String
+}
+
+type MockPushConfig struct {}
```
warpfork, line 456: For the purpose of getting it visible in this PR: here's copypaste of the (real!) example config from [the related branch in warpsys](https://github.com/warptools/warpsys/blob/a22c19624ea6e8902700e68e9e6cd912693a4d4d/.warpforge/mirroring.json):

```
{
	"ca+https://warpsys-wares.s3.fr-par.scw.cloud": {
		"pushConfig": {
			"s3": {
				"endpoint": "https://s3.fr-par.scw.cloud",
				"region": "fr-par",
				"bucket": "warpsys-wares"
			}
		}
	}
}
```

warpfork: I think we're probably missing a capsule type here.  Something that says what this data is, and gives us a position to put version indicators.

It's possible we might want to think about splitting some of this into another schema file too -- to group together all the types that are only used to describe values that appear only in workspace config.

wfapi/wfapi.ipldsch
```diff
@@ -430,4 +430,27 @@ type CatalogMirrorsByPacktype {Packtype:WarehouseAddrList}
 
 type ReplayCapsule union {
 	| Plot "plot.v1"
-} representation keyed
\ No newline at end of file
+} representation keyed
+
+# MirroringConfig defines the mirroring configuration for an entire workspace.
+# This maps the addresses which we want to publish to (of type WarehouseAddr) 
+# to specific configs, which may be backed by various types of storage.
```
warpfork, line 437: Could use some discussion about the prefixing logic, and some examples.

(There aren't really well established formatting rules for these comments, but I think using indentation for example content, roughly like godoc does, is probably fine and well.)

wfapi/wfapi.ipldsch
```diff
@@ -430,4 +430,27 @@ type CatalogMirrorsByPacktype {Packtype:WarehouseAddrList}
...
+type MirroringConfig {WarehouseAddr:WarehouseMirroringConfig}
+
+type WarehouseMirroringConfig struct {
+	pushConfig WarehousePushConfig 
+}
```
warpfork, line 442: Seems fine to put this here as room for expansion, but I'm curious if you've got any likely examples for future fields that might be added to this struct?  (It's not flunking any vibe checks for me, but I don't have concrete examples in my mind yet either, which seems, ehh.)

go.mod
TripleDogDare: You can somewhat get this number from the CI build artifact. If you want something more detailed, it should probably emit from the build CI job.

master | 5f3e22b7e86cf53be904bf36b8e51d4478fdf065: 23.7MB
branch | 93ec2c1a24cc3555eed1bfac22564ad911c51716:  26.4 MB

pkg/mirroring/s3.go
ericevenchick: Agree, missed this. Resolved now :)

pkg/mirroring/s3.go
ericevenchick: Cause I'm silly. Fixed.

pkg/mirroring/s3.go
ericevenchick: `wfapi.WareID.Subpath()` now does this.

pkg/mirroring/push.go
ericevenchick: Yeah, Go problems. I resolved it as suggested.

go.mod
ericevenchick: I think < 3 MB is sufficiently small that we don't care?

TripleDogDare: FYI that's tarball size. I didn't decompress it.

examples/500-cli/cli.md
ericevenchick: Yes, but now we only use that mirror file for testing mirroring which fixes the issue :)

go.mod
ericevenchick: Binary from `master` is 28M
Binary from this branch is 35M

ericevenchick: Yeah, we don't love that this adds to bin size, but for now we deemed this to be okay. Also note, we can save more than this adds with build flags to remove debug info!

warpfork (approved):
LGTM.

I wonder if we'll revisit where some of this code lives -- in separate processes, with serial APIs?  Closer to what we already currently have with rio? -- but for the present, putting it all in one place in preference for development velocity seems like a tradeoff that's reasonable for us to make.

PR #64, opened 2023-01-26
Misc Fixes (Remove Debug Println, WarePath, Logging)

ericevenchick:


ericevenchick:
This now fixes 3 things:
1. A stray `Println` in `GetRelease`
2. Move logging for "pushing ware" so it is only shown when the ware is actually pushed
3. Update `WarePath` to use the correct path (previously failed for `.warphome`) and update CachePath to use the `WareID.Subpath()` helper.

TripleDogDare (approved):
LGTM

PR #66, opened 2023-01-28
Make relative mounts relative to formula/module

ericevenchick:
With the shuffling of paths, we ended up making mounts relative to the working directory where `warpforge` is executed. This is probably not what we want, since we have Formulas that provide relative paths to the `formula.wf` (or `module.wf` / `plot.wf`) file.

This resolves that by adding a parameter to `FormulaExecConfig` / `PlotExecConfig`. If provided, this specifies the path to use for relative mounts (`FormulaDirectory`).

This also adds a test which uses the CLI testing. This seemed like the best way to build and test a filesystem.

### DIFF
--- a/pkg/formulaexec/formula_exec.go
+++ b/pkg/formulaexec/formula_exec.go
@@ -86,8 +86,11 @@ type ExecConfig struct {
 	RunPathBase string
 	// WarehousePathOverride overrides the directory where outputs are stored.
 	WhPathOverride *string
-	// WorkingDirectory will be where relative paths are based from.
+	// WorkingDirectory is the directory we are running warpforge from
 	WorkingDirectory string
+	// FormulaDirectory is the location of the formula (or module) being run
+	// Relative mount paths are relative to this path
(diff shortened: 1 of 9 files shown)


pkg/formulaexec/formula_exec.go
```diff
@@ -845,7 +848,7 @@ func execFormula(ctx context.Context, cfg internalConfig) (wfapi.RunRecord, erro
 					hostPath = inputSimple.Mount.HostPath
 				} else {
 					// otherwise, use relative path
-					hostPath = filepath.Join(cfg.WorkingDirectory, inputSimple.Mount.HostPath)
+					hostPath = filepath.Join(cfg.FormulaDirectory, inputSimple.Mount.HostPath)
```
TripleDogDare, line 851: Kind of surprised that this is the only line that changes in formula exec.

TripleDogDare:
Do we need tests for the `run ./...` incantation?

TripleDogDare (approved):
lgtm

PR #69, opened 2023-02-02
Ferk tweaks

ericevenchick:
Makes some changes to how Ferk works:
1. Allows the plot specified to be a plot file or a module directory
2. Modify ferk to run a specific step in the plot (default: `ferk`)

Most notably, this adds `dab.FindActionableFromFS` which finds actionable files (modules, formulas, plots) in a standard way. This is only use by `ferk` as of this PR, but will be integrated into other subcommands (notably, `run`) in the near future :)

### DIFF
--- a/cmd/warpforge/ferk.go
+++ b/cmd/warpforge/ferk.go
@@ -34,7 +34,6 @@ var ferkCmdDef = cli.Command{
 			Name:    "plot",
 			Aliases: []string{"p"},
 			Usage:   "Specify a plot file or module directory to use.  The current directory is used by default.  If the current directory isn't a module, or if this flag is explicitly set to empty string, a very minimal default plot with a simple base image will be used.",
-			Value:   ".",
 		},
 		&cli.StringFlag{
 			Name:    "step",
@@ -65,6 +64,10 @@ func cmdFerk(c *cli.Context) error {
 	var err error
 	ctx := c.Context
 	log := logging.Ctx(ctx)
+	cwd, err := os.Getwd()
+	if err != nil {
+		return serum.Errorf(wfapi.ECodeIo, "failed to get working directory")
+	}
 
 	plot, plotDir, err := cmdFerk_selectPlot(c)
 	if err != nil {
@@ -133,7 +136,11 @@ func cmdFerk(c *cli.Context) error {
 		},
 	}
 
-	exCfg, err := config.PlotExecConfig(&plotDir)
+	// if no plotDir was provided, use the working directory
+	if plotDir == nil {
+		plotDir = &cwd
+	}
+	exCfg, err := config.PlotExecConfig(plotDir)
 	if err != nil {
 		return err
 	}
@@ -161,15 +168,17 @@ func cmdFerk(c *cli.Context) error {
 	return nil
 }
 
-func cmdFerk_selectPlot(c *cli.Context) (plot *wfapi.Plot, plotDir string, err error) {
+func cmdFerk_selectPlot(c *cli.Context) (*wfapi.Plot, *string, error) {
 	// If emptystring plot: make a default one.
 	if c.String("plot") == "" {
-		_, err = ipld.Unmarshal([]byte(util.FerkPlotTemplate), json.Decode, &plot, wfapi.TypeSystem.TypeByName("Plot"))
+		plot := wfapi.Plot{}
+		_, err := ipld.Unmarshal([]byte(util.FerkPlotTemplate), json.Decode, &plot, wfapi.TypeSystem.TypeByName("Plot"))
 		if err != nil {
 			panic(err) // This is a fixed value at compile time.  It can't fail unless there's a bug.
 		}
-		return
+		return &plot, nil, nil
 	}
+	fmt.Println("plot is", c.String("plot"))
 
 	// Look for module and plot files.
 	// If we received a filename: expect that to be a plot.
@@ -177,18 +186,18 @@ func cmdFerk_selectPlot(c *cli.Context) (plot *wfapi.Plot, plotDir string, err e
 	// If it's a dir: look for a module, and then look for its plot.
 	pth, err := filepath.Abs(c.String("plot"))
 	if err != nil {
-		return
+		return nil, nil, err
 	}
 	m, p, f, _, _, err := dab.FindActionableFromFS(os.DirFS("/"), pth, "", false, dab.ActionableSearch_Any)
 	_, _ = m, f // TODO support these
 	if err != nil {
-		return nil, "", err
+		return nil, nil, err
 	}
 	if p == nil {
-		return nil, "", serum.Error(wfapi.ECodeMissing,
+		return nil, nil, serum.Error(wfapi.ECodeMissing,
 			serum.WithMessageTemplate("could not find a plot given path {{path|q}}"),
 			serum.WithDetail("path", pth),
 		)
 	}
-	return p, pth, err
+	return p, &pth, err
 }
--- a/cmd/warpforge/ferk.go
+++ b/cmd/warpforge/ferk.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/warptools/warpforge/cmd/warpforge/internal/util"
 	"github.com/warptools/warpforge/pkg/config"
+	"github.com/warptools/warpforge/pkg/dab"
 	"github.com/warptools/warpforge/pkg/logging"
 	"github.com/warptools/warpforge/pkg/plotexec"
 	"github.com/warptools/warpforge/pkg/workspace"
@@ -32,7 +33,14 @@ var ferkCmdDef = cli.Command{
 		&cli.StringFlag{
 			Name:    "plot",
 			Aliases: []string{"p"},
-			Usage:   "Specify a plot file to use.  If not set, a very minimal default plot with a simple base image will be used.",
+			Usage:   "Specify a plot file or module directory to use.  The current directory is used by default.  If the current directory isn't a module, or if this flag is explicitly set to empty string, a very minimal default plot with a simple base image will be used.",
+			Value:   ".",
+		},
+		&cli.StringFlag{
+			Name:    "step",
+			Aliases: []string{"s"},
+			Usage:   "Name a step in the plot that we want to run interactively.  Any other steps leading up to it will be evaluated noninteractively.",
+			Value:   "ferk",
 		},
 		&cli.StringFlag{
 			Name:  "rootfs",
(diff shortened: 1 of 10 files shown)


pkg/dab/find.go
```diff
@@ -0,0 +1,237 @@
...
+	"github.com/ipld/go-ipld-prime/codec/json"
+	"github.com/serum-errors/go-serum"
+	"github.com/warptools/warpforge/wfapi"
+)
+
+// ActionableSearch is a bitfield for what the ActionableFromFS function should treat as acceptable findings.
+type ActionableSearch uint8
+
+const (
+	ActionableSearch_None    ActionableSearch = 0
```
TripleDogDare, line 20: @warpfork and I discussed this very briefly and it may be more readable being explicit, but I still think it's fine to use ` = 1 << iota -1` here and let the compiler handle the flag setting by declaration order.

pkg/dab/guess.go
```diff
@@ -0,0 +1,80 @@
...
+//   - warpforge-error-serialization -- if none of the keywords (or, no strings at all)
+//      can be found within the scan zone.
+func GuessDocumentType(scanMe []byte, keywords []string) (string, error) {
+	// This code takes considerable shortcuts, and is specific to JSON and limited domains of data.
+	//
+	// Ideally, feature would use codecs, and take some kind of codec system as a parameter.
+	// However, a good codec API to accomplish this task efficiently must support token stepping,
+	// because we want to return after a few tokens rather than the whole stream
+	// (and perhaps to be really optimal, it would be nice to even be able to return a buffer of the tokens and a codec that can be resumed).
+	// Regretably, `codec ipld.Decoder` does not expose support for token stepping;
```
TripleDogDare, line 51:
```suggestion
	// Regrettably, `codec ipld.Decoder` does not expose support for token stepping;
```

cmd/warpforge/ferk.go
```diff
@@ -155,10 +160,44 @@ func cmdFerk(c *cli.Context) error {
...
+	// If emptystring plot: make a default one.
+	if c.String("plot") == "" {
+		plot := wfapi.Plot{}
+		_, err := ipld.Unmarshal([]byte(util.FerkPlotTemplate), json.Decode, &plot, wfapi.TypeSystem.TypeByName("Plot"))
+		if err != nil {
+			panic(err) // This is a fixed value at compile time.  It can't fail unless there's a bug.
+		}
+		return &plot, nil, nil
+	}
+	fmt.Println("plot is", c.String("plot"))
```
TripleDogDare, line 181: Is this supposed to be command output?

pkg/dab/find.go
```diff
@@ -0,0 +1,237 @@
...
+	// Stat it and consider that first:
+	// if it is a file, that's its own whole detection procedure (and means no further search);
+	// if it's not a file, we can procede to the behavior for dir feature detection (without popping).
+	foundPath = filepath.Join(basisPath, remainingSearchPath)
+	fi, e2 := stat(fsys, foundPath)
+	if e2 != nil {
+		err = e2
+		return
+	}
+	if fi.Mode()&^fs.ModePerm == 0 {
```
TripleDogDare, line 114: This logic feels ambiguous. `&^` is bitwise-AND _and_ bitwise-XOR? I don't know what this means.

pkg/dab/find.go
```diff
@@ -0,0 +1,237 @@
...
+// Errors:
+//
+//  - warpforge-error-datatoonew -- when found data is not supported by this version of warpforge
+//  - warpforge-error-invalid-argument -- when a file contains the wrong content for its filename
+//  - warpforge-error-io -- when an IO error occurs during search
+//  - warpforge-error-missing -- when an expected file is missing
+//  - warpforge-error-module-invalid -- when a read module contains invalid data
+//  - warpforge-error-searching-filesystem -- when the search of the filesystem produces an invalid result
+//  - warpforge-error-serialization -- when IPLD deserialization fails
+func FindActionableFromFS(
```
TripleDogDare, line 94: Needs tests. Probably a _lot_ of them.

pkg/dab/find.go
```diff
@@ -0,0 +1,237 @@
...
+			p, e2 = PlotFromFile(fsys, foundPath)
+			if serum.Code(e2) != wfapi.ECodeMissing { // notexist is ignored.
+				// Any error that's just just notexist: means our search has blind spots: error out.
+				err = wfapi.ErrorSearchingFilesystem("modules, plots, or formulas", e2)
+				return
+			}
+		}
+
+		// Peek for Formula.
+		// TODO
```
TripleDogDare, line 164: Without a description I assume this TODO was supposed to be implemented in this PR.

pkg/dab/find.go
```diff
@@ -0,0 +1,237 @@
...
+	// Iteratively check directories.
+	// In each directory, look for well-known file names.
+	// If we find a module, that dominates; then plot; then formula.
+	// When there's no match, if searchUp is enabled, then pop one segment off searchPath and search again.
+	for {
+		// Peek for module.
+		if accept&ActionableSearch_Module > 0 {
+			foundPath = filepath.Join(basisPath, remainingSearchPath, "module.wf")
+			m, e2 = ModuleFromFile(fsys, foundPath)
+			if serum.Code(e2) != wfapi.ECodeMissing { // notexist is ignored.
```
TripleDogDare, line 140: This probably resolves to `true` for a `nil` error (`Code` would return an empty string). I don't think this describes the expected behavior. Probably need tests for this.

TripleDogDare (dismissed):
Primarily `dab.FindActionableFromFS` needs tests. There's a lot of branches and it will have a lot of edge cases to get through.

pkg/dab/find.go
ericevenchick: I have no idea what this is doing, @warpfork wrote it... can you explain?

TripleDogDare: I looked it up and it's a bit clear (AND NOT) operation. Feels somewhat special and contrarian to lack a bitwise NOT operator. It's possible to do a bitwise NOT but it's an XOR NEGATE `(^-x)` so a bit clear would be `a ^ (^-b)` which... yeah bit clear needs to not be that. No idea why they didn't use `&~` but whatever.
```
package main

import "fmt"

func main() {
	a := 0b1010
	b := 0b0011
	fmt.Printf("       a = %#04b\n", a)
	fmt.Printf("       b = %#04b\n", b)
	fmt.Printf("   a ^ b = %#04b\n", a^b)
	fmt.Printf("   a & b = %#04b\n", a&b)
	fmt.Printf("  a &^ b = %#04b\n", a&^b)
	fmt.Printf("a ^(^-b) = %#04b\n", a^(^-b))
}
```
```
       a = 0b1010
       b = 0b0011
   a ^ b = 0b1001
   a & b = 0b0010
  a &^ b = 0b1000
a ^(^-b) = 0b1000

Program exited.
```
https://go.dev/play/p/3r_EypT_FHD

warpfork: Yeah this syntax surprised me to rediscover when I wrote it too, heh.

Overall what this means is "is it a normal file".  I wish there was a clearer way to write this (and if you can think of one, please go for it).

pkg/dab/find.go
warpfork: It's not reachable where this function gets used by ferk mode.  It won't come up until we replace the current "run" subcommand's search logic with this function (that one does look for and use formulas, whereas ferk as yet has not supported that).  Or if we expanding the features of ferk to support formulas.

Probably worth turning into a panic, though.

pkg/dab/find.go
TripleDogDare: Why not use `fi.Mode().IsRegular()`?

pkg/dab/find.go
TripleDogDare: Okay. Definitely want a panic and/or additional comments.

pkg/dab/find.go
warpfork: ... yep, that'd be better.  Dunno how I missed it.

TripleDogDare: Tests still open irregular files during search mode. I don't really want to think about that right now so :shrug:.

pkg/dab/find_test.go
```diff
@@ -0,0 +1,211 @@
...
+func TestFindActionableFromFS_NotFound(t *testing.T) {
+	t.Run("when no actionable files exist", func(t *testing.T) {
+		fsys := fstest.MapFS{}
+		m, p, f, path, rem, err := FindActionableFromFS(fsys, "", "non/existing/directory", true, ActionableSearch_Any)
+		qt.Assert(t, err, qt.IsNil)
+		qt.Assert(t, m, qt.IsNil)
+		qt.Assert(t, p, qt.IsNil)
+		qt.Assert(t, f, qt.IsNil)
+		qt.Assert(t, path, qt.Equals, "")
+		qt.Assert(t, rem, qt.Equals, "")
```
warpfork, line 49: Reading these test cases makes me wonder if there really should be a struct here.

Either a struct that we plonk all the function returns into (just so we can assert everything at once, and to help condense the tests into a table)....

... or this is really making me wonder if I was wrong for not just having this function return a struct in the first place.

pkg/dab/find.go
```diff
@@ -165,35 +172,78 @@ func FindActionableFromFS(
...
+		}
+		return nil, nil
+	}
+	return p, nil
+}
+
+const (
+	guessMagic_ModuleV1 = "module.v1"
+	guessMagic_PlotV1   = "plot.v1"
+	guessMagic_Formula  = "formula" // REVIEW: not sure why we don't use FormulaCapsule V1 value here
```
warpfork, line 224: Because there's no such thing.  We chose to put a version-hinting capsule type around Formula and around FormulaContext separately... and thus the top level object in a file containing FormulaAndContext doesn't actually have any version hinting in it: those hints are in deeper structures.

Whether all that was a good idea or not...  I really don't know.  (It's not an accident; we did think about it at the time, in some depth.  But look at it again now: it certainly does look Weird.)

pkg/dab/find_test.go
warpfork: Hm.  Reading further test cases below, now I see that having the function return a struct wouldn't solve both desires at once, because the later tests are using `qt.IsNotNil`.

It might be an option to make a struct of bools and have a helper function that reduces return value nillity to that struct of bools.  That would help with the condensing into a table.  Maybe still worth it?

pkg/dab/find.go
```diff
@@ -165,35 +172,78 @@ func FindActionableFromFS(
 
 		// None of the filename peeks found a thing?
 		// Okay.  If we can search further up, let's do so.
+		foundPath = "" // clear found path
 		if !searchUp {
-			foundPath = ""
+			remainingSearchPath = filepath.Dir(remainingSearchPath)
 			return
 		}
```
warpfork, line 179: Ahha!  The tests found some things.  Thank you :bow:

pkg/dab/find.go
```diff
@@ -99,6 +99,18 @@ func FindActionableFromFS(
...
+			// canonicalize to an empty string for an empty remaining search path
+			remainingSearchPath = ""
+		}
+	}()
+	if filepath.IsAbs(searchPath) {
+		err = serum.Error(wfapi.ECodeArgument,
+			serum.WithMessageTemplate("search path {{path|q}} may not be an absolute path"),
+			serum.WithDetail("path", searchPath),
+		)
+	}
```
warpfork, line 113: I think I'd be just as happy having a panic.  (My thinking is that this function is far enough away from direct user inputs that the path parameters probably should've been munged and checked for good form at least once already somewhere closer to the main method.)

But it's a weakly held opinion.  And would be easy to change later one way or the other.

warpfork (approved):
This looks pretty solid now.

I have one wishful request around tests getting a little more table-oriented, because the current assertions spread out across so many lines that I found it a bit hard to quickly assemble a mental model of what cases are covered vs perhaps not.  But it's a nice-to-have.

The tests are they and they do look like they cover a lot.  :D

pkg/dab/find_test.go
TripleDogDare: We can make condense some of it but doing tables with so many diverse inputs and outputs is going to be pretty rough looking regardless of what we do.

pkg/dab/find.go
TripleDogDare: There's still a `Capsule` type in the schema. https://github.com/warptools/warpforge/blob/master/wfapi/wfapi.ipldsch#L40-L43

PR #70, opened 2023-02-02
add tests and fixes for workspace detection.

TripleDogDare:
This includes a test which finds an edge case which will not halt when executing the old FindWorkspace functions.
The goal here is to exercise the FindWorkspace functions thoroughly so we don't find too many more surprises.

This is split off from #58 to reduce PR size.


ericevenchick (approved):
lgtm, many error improvements 👍

PR #71, opened 2023-02-06
allow rio to be mounted via symbolic links

### LINKED ISSUE #68: warpforge fails if `rio` is a symlink
I'm currently packaging warpforge for [Nix](https://nixos.org/) and am compiling `rio` as a separate package.
In the Nix ecosystem, it's idiomatic to use symlinks for interop of packages, so my go-to solution of packaging warpforge would mean that `rio` is symlinked into `WARPFORGE_PATH`, instead of copied.
However, trying to execute a plot results in a ton of repeated error messages, essentially boiling down to:
```
the invoke runc for rio pack of output failed engine reported error: warpforge-error-executor-failed: executor engine failed: the runc engine reported error:  time="2023-02-02T15:53:09+01:00" level=error msg="runc run failed: unable to start container process: exec: \"/.warpforge.container/bin/rio\": stat /.warpforge.container/bin/rio: no such file or directory"
```

Digging into the code a bit, it seems like the plugin directory is bindmounted into the `runc` containers, and it seems like mounted symlinks are a general pain point of runc-based containers, as indicated by e.g. [this stackoverflow question](https://stackoverflow.com/questions/38485607/mount-host-directory-with-a-symbolic-link-inside-in-docker-container).

I know that this is not the intended/official way of installation, but it would be great if symlinked plugins could still be supported, or if `warpforge status` and/or `warpforge health` could complain about symlinks otherwise.

TripleDogDare:
This requires following the symbolic links from the binary path. However, if a source or destination path is repeated we should not mount it twice. This de-duplication might need some tweaking.

I'm not sure if mounting an entire chain of symbolic links is a _good_ idea since you may be mounting unexpected things, but we're already doing that to some extent. For example, I can link `$HOME/gopath/bin/rio` with an absolute path link to `$HOME/repos/warpforge/plugins/rio` and it will mount the directory `$HOME/repos/warpforge/plugins` _inside_ the container to satisfy the link.

Testing this is... not easy. We would need both an `os.DirFS` and a `MapFS` with a `Stat` implementation which doesn't follow links like `Lstat`. Or we can project onto an actual filesystem and run tests from there which is a pain.

Fixes #68 .

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,8 @@ MODULE := $(shell go list -m)
 install:
 	@echo "Installing plugins..."
 	mkdir -p $(GOBIN)
-	cp ./plugins/* $(GOBIN)
+# ignore errors when copying plugins (cp does not like copying over symbolic links)
+	-cp ./plugins/* $(GOBIN)
 	@echo "Building and installing warpforge..."
 	go install ./...
 	@echo "Install complete!"
--- a/pkg/formulaexec/formula_exec.go
+++ b/pkg/formulaexec/formula_exec.go
@@ -259,45 +259,46 @@ func isSymbolicLink(m fs.FileMode) bool {
 	return m&fs.ModeSymlink == fs.ModeSymlink
 }
 
-func getBinMounts(binPath string) []specs.Mount {
+func getBinMounts(ctx context.Context, binPath string) []specs.Mount {
 	wfBinMount := specs.Mount{
 		Source:      binPath,
 		Destination: containerBinPath(),
 		Type:        "none",
 		Options:     []string{"rbind", "ro"},
 	}
+	log := logging.Ctx(ctx)
 	result := []specs.Mount{wfBinMount}
-	binFiles, err := ioutil.ReadDir(binPath) // note that this doesn't support nested directories.
+	rioPath := filepath.Join(binPath, "rio")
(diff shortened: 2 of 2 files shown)


warpfork:
I think in a situation like this where there's exactly the one file that needs to be mounted into place, it's probably better to do a single bind mount for that purpose, with the fully resolved symlink target.  I don't really see any reason to do a whole bunch of mounts and put the whole set of symlinks into the prep container just to let it resolve them again -- it's not necessary; and as you say, increases the surface area of what's visible in the container, which is not necessarily desirable, and likely to be unexpected.

Even if we had some plugin to handle which required a whole directory (perhaps because a plugin was something that was shipped as a Zapp?), then tracing around intermediate symlink hops is still unlikely to do anything of value.  We'd have to know to bring the relevant directory in; and the intermediate hops still wouldn't matter.

Testingwise: this would need a full integration/smoke test.  When mounting is involved, there's no way around that.

TripleDogDare:
That comment about testing was specifically just for "does this create the expected `[]specs.Mount`" more so than "actually test that the things work" which definitely does require a smoke test.

warpfork (approved):
Considering the comment in the issue that says it this solved the problem, LGTM!

PR #76, opened 2023-06-17
cli redecorating

warpfork:
This set of commits changes the way the helptext is generated for the CLI.

The output is now markdownlike (or, exactly markdown, if called in certain ways), and colorized in the terminal using ANSI escape sequences.  Flags galore for this.

The overall text is a bit less compact in most cases, but I think nicely legible.

I've been driving this around on my localhost for a while and grown to like it.

The commit history contains a variety of experiments.  Most have been removed again by the end.  In general, this was a process of trying something that seemed to have nice bells and whistles, and then finding some glaring issue, and then replacing it with something simpler (admittedly at the cost of taking on more code ourselves).  So, some initial attempts to use all-in-one rendering libraries ended up replaced with doing our own goldmark hooks directly.

On the down sides: This heap of code does also raise the complexity of the CLI rigging quite a bit.  (This turned out to be a lot more complicated to glue together than I expected at the outset!)  Some of that is now fully paid already; some might become ongoing.  The most worrying part in terms of maintainability is that slightly different codepaths end up exercised if you use markdown mode and colored markdown-ish mode.

I wrote some test infrastructure to store the markdown help text into the website repo, which seems like it could produce good docs.  Having subcommands link to each other when referenced in the docs is still a wishlist item for the future, but should be supportable within the work done here.


warpfork:
After letting this steep for a while:

- I don't love this, honestly.  The maintainability cost step up is... high.
- But if we _do_ ditch this... and then try to do _any_ tweaking to the CLI output at all... we'll end up in the same complexity threshhold again _rapidly_.

The latter is my main argument for merging it.  Detangling all this template stuff in the CLI library is unavoidable if we want to tweak it at all.  And also the whole markdown parse phase actually frees us from insanity with the golang templating whitespace issues.  It's mindboggling that this seems like the "simplest" way forward, but... it kinda does.

PR #80, opened 2023-09-11
Simple "server" implementation for catalog.

TripleDogDare:
Generates the html and serves. It requires a different url-prefix for links so beware of that. We also need to think about how to handle mirrors and download links more effectively.

### DIFF
--- a/app/_docs/warpforge-catalog.md
+++ b/app/_docs/warpforge-catalog.md
@@ -49,6 +49,9 @@ Generates HTML output for the root workspace catalog containing information on m
 ### mirror
 Mirror the contents of a catalog to remote warehouses
 
+### serve
+Generates html and serves
+
 ### help, h
 Shows a list of commands or help for one command
 
--- a/app/catalog/catalog_cli.go
+++ b/app/catalog/catalog_cli.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"net/http"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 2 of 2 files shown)


app/catalog/catalog_cli.go
```diff
@@ -715,69 +747,116 @@ func cmdCatalogShow(c *cli.Context) error {
...
+	}
+	os.RemoveAll(cfg.OutputPath)
+	if err := cfg.CatalogAndChildrenToHtml(); err != nil {
+		return fmt.Errorf("failed to generate html: %s", err)
+	}
+
+	fmt.Printf("published HTML for catalog %q to %s\n", pcfg.catalogName, cfg.OutputPath)
+
+	dirHandler := http.Dir(cfg.OutputPath)
+	fmt.Printf("Serving at http://%s\n", addr)
```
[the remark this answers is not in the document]
warpfork: Adding the actual feature behind that should be just a mux, a few lines below here, roughly:

```
mux := NewServeMux()
mux.Handle("/", http.Dir(cfg.OutputPath))
mux.Handle("/_wares/", http.Dir(workspaceWareDir)
```

(well, and then get the workspace info down here, so we actually have that path.)

app/catalog/catalog_cli.go
```diff
@@ -156,6 +157,37 @@ var catalogCmdDef = &cli.Command{
...
+				util.CmdMiddlewareLogging,
+				util.CmdMiddlewareTracingConfig,
+				util.CmdMiddlewareTracingSpan,
+			),
+			Flags: []cli.Flag{
+				&cli.StringFlag{
+					Name:    "output",
+					Aliases: []string{"o"},
+					Usage:   "Output path for HTML generation",
+				},
```
warpfork, line 173: Might prefer to get rid of this, per other comments (below) about making this as works-by-default as possible.  We can just pick a path under `{rootWorkspaceControlDir}/tmp/` (or `$WARPFORGE_RUNPATH`?)

app/catalog/catalog_cli.go
```diff
@@ -156,6 +157,37 @@ var catalogCmdDef = &cli.Command{
...
+				&cli.StringFlag{
+					Name:  "url-prefix",
+					Usage: "URL prefix for links within generated HTML",
+					Value: "/",
+				},
```
warpfork, line 178: Let's drop this too, per comment below about making this as works-by-default as possible.  We control the whole mux that's processing these URLs and not supporting configuring it, so this option doesn't even make sense in that story.

app/_docs/warpforge-catalog.md
```diff
@@ -49,6 +49,9 @@ Generates HTML output for the root workspace catalog containing information on m
 ### mirror
 Mirror the contents of a catalog to remote warehouses
 
+### serve
+Generates html and serves
```
warpfork, line 53: Propose:

> Serve a local http site that contains a navigable copy of the Catalog, and hosts a warehouse at `_wares/`.  (Meant for rapid sharing in LAN environments.)

warpfork (commented):
I like where this is going :)

PR #83, opened 2023-10-02
Loosens restrictions on catalog ls results.

TripleDogDare:
Removes the IsDir check for entries in the catalogs directory when returning the list of catalog names.
This allows catalogs to be symbolic links (or something else) and still work as long as it resolves properly.


warpfork:
Yep, makes sense to me symlinks should work there.  LGTM.

warpfork (approved):

